PR #5, opened 2025-05-12
changed Control-f to Ctrl+f

JaredTweed:
pretty simple change, just thought this would be more standard


kolunmi:
Thank you for your contribution but this is just a placeholder label, I'm planning to get rid of it

PR #11, opened 2025-05-18
feat: Add initial devcontainer support

AtomHare:
This PR adds initial support to develop bazaar through a devcontainer. This allows easier contributions and dev environment from immutable distros like bluefin. 

The dockerfile is required because ubuntu 24.04 doesn't have all the required libs to build it. Hence, ubuntu 25.04 is used.

It needs more work around flatpak to either bind to the host flatpak or make flatpak work properly inside the container to allow testing. Since wayland passthrough is enabled it's possible to launch bazaar as expected through the README after building it.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:25.04
+
+RUN userdel -f -r ubuntu
+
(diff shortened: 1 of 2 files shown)


kolunmi:
interesting, thank you for your contribution

PR #17, opened 2025-05-19
feat: add RPM package

tulilirockz:
We are already packaging this on [ublue-os/packages](https://github.com/ublue-os/packages) but I thought it would be best to upstream this too

You can test the build on a Fedora 42 container then installing `rpmdevtools` and `dnf-plugins-core`.

```bash
dnf builddep -y ./bazaar.spec
spectool -agR ./bazaar.spec
rpmbuild -bb ./bazaar.spec
```


kolunmi:
sweet, thanks!

PR #18, opened 2025-05-20
fix: desktop file not launching on the flatpak/rpm versions

tulilirockz:
Signed-off-by: Tulip Blossom <tulilirockz@proton.me>


kolunmi:
Thanks! :)

PR #20, opened 2025-05-22
feat: add justfile for building

tulilirockz:
Just a few helpers for building the flatpak/rpm/base stuff, I was going to make a PR for the automated workflows but I'm tired today so here is this!


kolunmi:
Thanks a lot! I'm tired too 😴

PR #27, opened 2025-06-02
Made clear how and which dependencies need to be installed

JaredTweed:
Just thought this would make local downloading easier, thus hopefully increasing contributions.


kolunmi:
Hey, thanks for the pr! Looking at the packages being installed, there are some odd ones like `gcc-c++` and `cmake`. These are not  dependencies of bazaar, so I'm not sure why you included them. Either way, these sorts of lists are specific to particular distros and will need to be maintained as the project changes, so I'm not interested in including them at the moment.

PR #28, opened 2025-06-07
chore(ci): add release please action

castrojo:
This automates releases for you, though you may need to do a manual release first since I don't think it goes back.

You need to make a token in github named: `MY_RELEASE_PLEASE_TOKEN` and ensure the right permissions are set so the action can release: 

https://github.com/googleapis/release-please-action#github-credentials


kolunmi:
Thanks, I believe everything is setup now

PR #91, opened 2025-07-07
docs: update readme

castrojo:
Some minor tweaks, the config file is more to pitch to distributions than end users so I tweaked that.


kolunmi:
Thanks!

PR #102, opened 2025-07-08
chore(l10n): update French translation

HarmfulBreeze:
This commit updates and improves the French translation by using more common words and sentence structures, and makes it more accurate to the original material.


HarmfulBreeze:
The Conventional Commit scope of other localization-related commits looks inconsistent, hopefully mine is okay?
Please let me know if you need me to adjust the commit title.

po/fr.po
```diff
@@ -26,28 +26,27 @@ msgstr "Bazaar"
 
 #: data/io.github.kolunmi.Bazaar.desktop.in:8
 msgid "GTK;System;PackageManager;Discover;Flatpak;Software;Store;"
-msgstr "GTK;Système;GestionnaireDePaquets;Discover;Flatpak;Logiciel;Store;"
+msgstr ""
+"GTK;Système;Gestionnaire de paquets;Découvrir;Flatpak;Logiciel;Magasin;"
```
AtomHare, line 30: Here Discover does not need to be translated since it's Discover as the KDE app store

po/fr.po
```diff
@@ -8,16 +8,16 @@ msgid ""
...
-"Last-Translator: Atom <29772841+AtomHare@users.noreply.github.com>\n"
+"POT-Creation-Date: 2025-07-08 14:12+0200\n"
+"PO-Revision-Date: 2025-07-08 15:13+0200\n"
+"Last-Translator: Léane GRASSER <leane.grasser@proton.me>\n"
 "Language-Team: French\n"
 "Language: fr\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n > 1);\n"
```
AtomHare, line 19: I think this line needs to be updated because it creates newlines starting with "", I'm not familiar enough with localization editors to know if it's something to take care of or not

po/fr.po
```diff
@@ -105,27 +105,27 @@ msgstr "Applications de la semaine"
...
 
 #: src/bz-flathub-page.blp:117
 msgid "Recently Updated"
-msgstr "Mises à jour"
+msgstr "Mises à jour récemment"
 
 #: src/bz-flathub-page.blp:149
 msgid "Recently Added"
-msgstr "Nouveautés"
+msgstr "Ajoutées récemment"
```
AtomHare, line 116: I took the same wording as the flathub french page to make Bazaar match (see https://flathub.org/fr?category=trending), @kolunmi what are your thoughts on this ?

po/fr.po
```diff
@@ -302,66 +304,67 @@ msgid ""
...
 #, c-format
 msgid "Additionally, %d runtimes and/or addons will be updated."
-msgstr "De plus, %d runtimes et/ou addons seront mis à jour"
+msgstr ""
+"De plus, %d environnements d'exécution et/ou extensions seront mis à jour."
 
 #: src/bz-window.blp:36
 msgid "Transactions Will Appear Here"
-msgstr "Les opérations apparaitrons ici"
+msgstr "Les transactions apparaîtront ici"
```
AtomHare, line 327: Here we talk about installating, removing or updating a package, to me it makes more sense to talk of this as "opérations" instead of "transactions" that are more monetary related in french

"une transaction" feels weird translated like that

AtomHare (commented):
Thank you for this translation update! I came up quickly with the previous one and made some errors/typos...

Since I'm not too much familiar with localizations, I raised some concerns in the comments.

And I reviewed the PR it looks good to me !

po/fr.po
kolunmi: @AtomHare whichever is the better fit in French, I trust your judgement

po/fr.po
HarmfulBreeze: I believe "transaction" is the technical term used by Flatpak [1] for these operations.
I think it's up to the dev to tell us whether we should stick with the technical term in our translations or update the source string to use the more user-friendly "operation" :)

[1] https://github.com/flatpak/flatpak/blob/0152272d6caf2622536fad8869573a76001a493b/common/flatpak-transaction.c#L65-L66

HarmfulBreeze:
Thanks for the review! I replied to your comments.

po/fr.po
kolunmi: stick to the more user-friendly term

kolunmi:
ready to merge this whenever you guys are ready

HarmfulBreeze:
If you're fine with keeping "Recently Added", this PR can be merged.

PR #103, opened 2025-07-08
Create pt-BR.po

leosgarcia:
Brazilian Portuguese translations for bazaar package.


kolunmi:
> You may not use llms to generate the strings (I could do that). If you do, I will ban you from the project

PR #108, opened 2025-07-08
feat(po): Add initial German Translation

renner0e:
EDIT: read everything again, should be good to go now


Mandarinoooo:
So, before I get started, I'm on bluefin-stable and thus running the latest Bazaar provided there. I may be missing context on a few translations.

L35: "animierten Symbolhintergrund auf der Startseite anzeigen" (supposing this is a toggle) would be more consistent with L31
L61:  As jorge always preaches, "Anwendung" would be preferable for normie users imo
L74: Torn here, but I'd say "Keine kuratierten Anwendungen" 
L81: ^^
L90: I most often see "Ein Fehler ist aufgetreten", both are fine (unless you're minmaxing on words)
L106: "Im Trend"
L211: maaaaybe "Tippen Sie hier", but at the same time, people aren't that stupid. I think.
L268: "In %.02f Sekunden abgeschlossen"?
L312: "(...) haben neue Aktualisierungen erhalten."? Would diverge from the original, but seems more understandable to me.
L374: "Kuratierte Anwendungen anzeigen [alternatively: ansehen, for coherence]" see L74
378: "Kuratiert" ...it kinda fits less well here. 
L394: "Seitenleiste für Transaktionen anzeigen"? I think the toggling aspect is intuitive enough
L410: Kinda sounds like the application is scolding you lmao, maybe "Das geht gerade nicht."
L426: "Die folgende Anwendung" see L61
L431: I believe "Sind sie sich sicher" would be more correct

Also, transactions is really too tech-y in my opinion. Shows especially in already very verbose german.
That's about it.

kolunmi:
Thank you!

PR #126, opened 2025-07-09
chore(po): add myself to credits to ru

Lumaeris:


kolunmi:
We're going to do this a different way

PR #127, opened 2025-07-09
chore(po): Minor tweaks to the Czech translation.

ExistingPerson08:
To make it similar to the translation in other Gnome applications + fix typos.


kolunmi:
Thanks!

PR #177, opened 2025-07-15
added pt_PT translations

azenyr:
Portuguese (from Portugal) is my native language and since on the Linux world we usually have a notorious lack of pt_PT translators, I decided to translate Bazaar myself. First time doing this (both translating an app and doing the github PR), so I hope everything was done right. Tell me if something needs to be adjusted. I used the commands and instructions provided in TRANSLATORS.md .

Please note that some DEs or distros configure the locale code "pt" instead of pt_PT, even KDE had a bug for some time (now fixed) that you could not select Portuguese from Portugal from the language settings UI since the options were configured in the backend as "pt_BR" and "pt" (not pt_PT) and it ended up having all apps showing Brazilian translations on both options. On this PR I used pt_PT since it is the correct Portuguese (Portugal) language code, even though some distros like to wrongly refer to it as just "pt". If Bazaar ends up getting just "pt" from the distro locale variables, please serve the correct pt_PT language (since pt_BR is usually well coded and defined by itself, as it already is here on Bazaar) and not Brazilian, since both languages are actually very different in many aspects and some sentences and words don't make much sense between both. Things like these happen because, again, the Linux world has a lack of Portuguese (Portugal) maintainers/translators so devs don't usually notice things like these.

TL;DR - If LC_MESSAGES=pt (not pt_PT, nor pt_BR), Bazaar should serve pt_PT by default, since thats how some (especially smaller) distros/DEs implement these two language locales. Most of modern DEs have the correct LC_MESSAGES=pt_PT nowdays, but still happens.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -8,6 +8,7 @@ fr
 id
 pl
 pt_BR
+pt_PT
 ru
 zh_CN
 zh_TW
--- a/po/pt_PT.po
+++ b/po/pt_PT.po
@@ -1,0 +1,587 @@
+# Portuguese translations for bazaar package
+# Tradução portuguesa (Portugal) para o pacote bazaar.
+# Copyright (C) 2025 THE bazaar'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the bazaar package.
+# Automatically generated, 2025.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: bazaar\n"
+"Report-Msgid-Bugs-To: \n"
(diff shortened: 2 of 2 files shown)


Guilherme-Enes:
Didn't notice someone else had translated as well to European Portuguese and did a pull request of my translation just now ... :sob: :sweat:

azenyr:
> Didn't notice someone else had translated as well to European Portuguese and did a pull request of my translation just now ... 😭 😓

Ooff sorry, you can compare both and tell me if you think you have suggestions to improve my translation in any way :)

kolunmi:
Thank you

Issue #181, opened 2025-07-15
app name in AppStream is not loaded correctly in non-English environments


jiesou:
In English environment, there are no problems:
<img width="2684" height="1744" alt="Image" src="https://github.com/user-attachments/assets/29bc8af2-ca8e-4933-8253-d8bbf8a344f4" />

But in Chinese environment (`LANGUAGE=zh_CN bazaar window --auto-service`)

<img width="2684" height="1744" alt="Image" src="https://github.com/user-attachments/assets/6175cab5-d3c4-4af0-9528-4286764f3952" />

Some application names fallback to id, according to the logic here:
https://github.com/kolunmi/bazaar/blob/ea31ad49469aa7af7de7b13e5d2d9a71d5541021/src/bz-flatpak-entry.c#L404-L406

According to the design of AppStream API, AppStream API should at least fallback to an app name with locale **C** , but here `as_component_get_name` returns null
https://github.com/ximion/appstream/blob/f1e84accbe535270f86e0633d39b0d77613fd549/src/as-component.c#L1377-L1382
https://github.com/ximion/appstream/blob/f1e84accbe535270f86e0633d39b0d77613fd549/src/as-context.c#L505

Really strange.

---

Base on my test:

normal - https://github.com/pleromix/IceBox/blob/main/io.github.pleromix.IceBox.metainfo.xml
normal - https://github.com/fabrialberio/Pins/blob/master/data/io.github.fabrialberio.pinapp.metainfo.xml.in
normal - https://codeberg.org/petsoi/words/src/branch/main/data/page.codeberg.petsoi.words.metainfo.xml.in.in

wrong - https://gitlab.gnome.org/GNOME/gnome-sudoku/-/blob/main/data/org.gnome.Sudoku.metainfo.xml.in.in
It falls back to the id, it does not have a zh-CN translation
wrong - https://github.com/AlexanderVanhee/Gradia/blob/master/data/be.alexandervanhee.gradia.metainfo.xml.in.in
It even has a zh-CN description translation: https://github.com/AlexanderVanhee/Gradia/blob/master/po/zh_CN.po
The app description are also correct, but the name is incorrect, falls back to the id
<img width="2684" height="1744" alt="Image" src="https://github.com/user-attachments/assets/7f78b2c6-5d24-4e05-8856-b25330a477fd" />

I can't find any pattern

jiesou:
This is a very deep problem.
[...]

PR #183, opened 2025-07-15
Fix app name in AppStream is not loaded correctly in non-English environments

jiesou:
Fixed: https://github.com/kolunmi/bazaar/issues/181

After cleaning up, there was only one small patch

The original XB_BUILDER_COMPILE_FLAG_SINGLE_LANG flag will cause all `lang` parameters in xml to be deleted, for example
`<name xml:lang="zh_CN">GNOME 数独</name>`

I've been understanding why `as_component_get_name_table` can't get anything except lang=C :sob:


kolunmi:
Thank you!

PR #194, opened 2025-07-19
fix(dbus): make dbus service have right binary path

tulilirockz:
Bazaar was writing this as `/bin/app` or something before, now it should be outputting the proper path on the dbus service. Makes it so the flatpak search provider can work completely fine by default


kolunmi:
Thank you

PR #199, opened 2025-07-20
fix: add thousands separators to download counts with locale support

ledif:
This PR uses the `%'d` specifier to automatically format numbers with appropriate separators based on user's locale settings.

## Before
Screenshot_20250720_125557

## After
Screenshot_20250720_125126


kolunmi:
Nice, thank you

PR #201, opened 2025-07-21
Feat: build Flaptak in ci

hanthor:


Justfile
```diff
@@ -60,3 +61,29 @@ build-rpm:
     dnf builddep -y "${RPMDIR}"/bazaar.spec
     rpmbuild -bb "${RPMDIR}"/bazaar.spec
     EOF
+
+[private]
```
salim-b, line 65: This must be the very first recipe to be the default, cf. https://just.systems/man/en/the-default-recipe.html

BTW, you can simply name it `_default` to be hidden to avoid the sep `[private]` attribute.

kolunmi:
Thank you

PR #217, opened 2025-08-12
flatpak: fix libdex version

tulilirockz:
We've been getting "connectivity" errors because its trying to use submodules without a network connection


kolunmi:
Thank you

PR #222, opened 2025-08-14
New translation: Ukrainian

Saltane:
Hi.

This PR does exactly what title suggests.

Note: names transliteration was ommited. I am not proficient enough at that, and it seems to be the usual case across GNOME apps anyway. Translated the app's title though, seems to be a common case too, and the change is as straightforward as it can get.

Translation tested on Bazzite, Bazaar version `0.2.3 no-git-info`.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -10,5 +10,6 @@ pl
 pt_BR
 pt_PT
 ru
+uk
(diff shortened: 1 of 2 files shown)


kolunmi:
Thank you very much

PR #227, opened 2025-08-17
chore(l10n): update Ukrainian translation

Saltane:
Still unsure what would be a better workflow — to stick with per-tag updates or with whenever there are new lines in main branch.


Lumaeris:
Whenever there are new lines in main branch for sure, kolunmi will add new strings when they're ready

PR #238, opened 2025-08-20
Improve and update pt_PT translation

azenyr:
This PR merges my improvement to pt_PT.po translation file. It contains improvements to existing translations (I noticed some small errors when using the pt_PT Bazaar app on my PC, and places where the message strings could be improved to better reflect the place where they are shown) as well as new translation strings that were added upstream by the app devs.


kolunmi:
Thank you!

PR #245, opened 2025-08-21
Update Polish translation

Micro856:


Micro856:
Closed it, because I forgot to edit one thing.

PR #261, opened 2025-08-23
fix: launching applications from flatpak-spawn

### LINKED ISSUE #247: Opening X11 apps from Flatpak fails silently
I'm testing out the Flatpak build right now and it seems to work fine, but a glaring issue I've found is that opening X11 only apps doesn't work, as I presume the display env var isn't passed in.

https://github.com/user-attachments/assets/99772119-ac53-47c6-b189-87ab2016e8dd

ilya-zlobintsev:
Fixes #247 
Previously opening installed programs (X11 or Wayland) largely didn't work, as the flatpak-system-helper service that runs the commands invoked by `flatpak-spawn` does not have a proper graphical environment. 
Using `systemd-run` resolves this, as it connects to the user's graphical session and has the correct environment.
Example of the environment difference:
```
> flatpak run --command=sh io.github.ilya_zlobintsev.LACT
F: Not sharing "/usr/share/icons" with sandbox: Path "/usr" is reserved by Flatpak
[📦 io.github.ilya_zlobintsev.LACT ~]$ flatpak-spawn --host sh -c 'echo $WAYLAND_DISPLAY'

[📦 io.github.ilya_zlobintsev.LACT ~]$ flatpak-spawn --host systemd-run --user --pipe sh -c 'echo $WAYLAND_DISPLAY'
Running as unit: run-p63500-i63501.service
wayland-0
```
With this PR both Wayland and X11 applications now launch correctly.

Obviously this won't work on systems without systemd, but I'm not sure if this is a major concern or not.

### DIFF
--- a/src/bz-flatpak-entry.c
+++ b/src/bz-flatpak-entry.c
@@ -938,7 +938,11 @@ bz_flatpak_entry_launch (BzFlatpakEntry    *self,
 
 #ifdef SANDBOXED_LIBFLATPAK
   fmt     = flatpak_ref_format_ref (FLATPAK_REF (ref));
-  cmdline = g_strdup_printf ("flatpak-spawn --host systemd-run --user --pipe flatpak run %s", fmt);
+
+  if (g_file_test ("/run/systemd", G_FILE_TEST_EXISTS))
+    cmdline = g_strdup_printf ("flatpak-spawn --host systemd-run --user --pipe flatpak run %s", fmt);
+  else
+    cmdline = g_strdup_printf ("flatpak-spawn --host flatpak run %s", fmt);
 
   return g_spawn_command_line_async (cmdline, error);
 #else


kolunmi:
Hi, thanks for the pr! Is there perhaps an environment variable we
could read which indicates that systemd is present on the system? We
could use that to adjust the behavior for systems without systemd.

ilya-zlobintsev:
It's possible to check if `/run/systemd` exists to determine if the system has systemd. What should the behaviour be if it doesn't? Should it throw an error?

kolunmi:
Hmm, let's just keep the old behavior if systemd doesn't exist. That way it has at least a chance at working

You can use the `g_file_test` function here, or I can edit the pr if you want

ilya-zlobintsev:
Done, though I don't have a non-systemd system to test if the fallback works.

kolunmi:
Looks great, thank you!

kolunmi:
sorry, still figuring out a new git client hehe

Novantric:
Thanks for fixing my issue!

PR #275, opened 2025-08-27
Update glycin dependency for glycin-2

KyleGospo:
Needed for Fedora rawhide, note that this will break building against older glycin. I would wait until glycin-2 is stable before merging this to make the transition as easy as possible.


alatiera:
Also need to update the bundled glycin in the flatpak manifest

kolunmi:
successfully updated to glycin-2!!!

PR #302, opened 2025-09-02

Saltane:
*Wake the frick up, Samurai. We have a string to translate* :dark_sunglasses:


kolunmi:
:crossed_swords:

PR #306, opened 2025-09-02
Rework button area on full view

AlexanderVanhee:
This PR reworks the button layout on the app listings page to reduce visual clutter. The previous design had 3 rows of pill buttons which looked awkward in my eyes.

Changes made:
- Moved the support button underneath the developer name with custom styling to make it more prominent, to be in line with the apps stated  goal of supporting app developers
- Relocated download statistics to the "App Details" section, which also now more obviously clarifies that the number shown represents downloads in the last month
- Changed the Share button to be styled flat to reduce the number of pill buttons

I would personally also suggest removing the install button when the app is already installed to reduce confusion, but I left it alone for now as there is probably a reason for it to be there.

Feel free once again to ask for changes.

Screenshot From 2025-09-02 14-35-22

image

### DIFF
--- a/src/bz-full-view.blp
+++ b/src/bz-full-view.blp
@@ -128,6 +128,25 @@ template $BzFullView: Adw.Bin {
                             };
                           }
                         }
+
+                        Button {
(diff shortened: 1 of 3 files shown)


kolunmi:
Phenomenal work. Thanks again!

PR #317, opened 2025-09-06
Move forge stars next to donation button

AlexanderVanhee:
This PR moves the Git Forge Star counter next to the support button and gives it a custom style.

Since giving a star is also a form of support, I think it makes sense to place it next to the support button. This change also helps reduce clutter in the main row.

image


kolunmi:
Awesome, thanks!

kolunmi:
that animation is really cool!

PR #323, opened 2025-09-07
Use relative colors to set Flathub colors

TheEvilSkeleton:
This also moves the Flathub style to `style.css`, to have it apply on light styles.


kolunmi:
Thank youuuu

PR #327, opened 2025-09-07
Rework the Installed page

AlexanderVanhee:
I made some adjustments that should hopefully lead to a better UX:

- Switched to Remove and Run buttons as i believe that most people visiting this page want to probably remove some app

- Set the application ID as the tooltip for the app name and removed the unique ID tooltip, since I don’t think even most developers need to worry about that.

- Moved the support button next to the app name and gave it styling similar to what you see on the app store page.

- Hid the permissions button under the “More actions” menu, as it’s not yet implemented. ( At first I thought I needed to install Flatseal for it to work, but the code indicates it’s simply not implemented at all.)

- Removed the “No support” button, since most users will probably understand that not having a donation link doesn’t necessarily mean the project can’t be supported. This also saves new translators some work.

Feel free to ask me to make further changes or bring things back.

<p align="center">
  image
</p>

<p align="center">
  image
</p>

### DIFF
--- a/src/bz-full-view.blp
+++ b/src/bz-full-view.blp
@@ -133,9 +133,9 @@ template $BzFullView: Adw.Bin {
                           Button {
                             visible: bind $invert_boolean($is_null(template.ui-entry as <$BzResult>.object as <$BzEntry>.donation-url) as <bool>) as <bool>;
                             styles [
-                              "pill",
+                              "small-pill",
                               "suggested-action",
-                              "support-button"
+                              "support"
                             ]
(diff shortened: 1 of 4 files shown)


kolunmi:
Wow thanks so much! Checking this out now.

By the way, do you happen to have an email I can contact you at? No pressure whatsoever!

kolunmi:
Looked awesome as usual, thanks again!
(btw my email is on my gh profile if you would prefer to contact me first!)

PR #330, opened 2025-09-07
app: Always start the application if no arguments are passed

alatiera:
When executing the binary, we should be starting the application
and service (if they are not already running) rather than telling
the user to run a different command.

This also avoids the need for the custom bazaar-launcher script


kolunmi:
Thanks for this PR, I will take a closer look in a bit!

kolunmi:
Thanks, I will build off of this :)

Sorry for the delay

PR #339, opened 2025-09-08

Saltane:
I should probably learn how to do a containerized setup.

### DIFF
--- a/po/uk.po
+++ b/po/uk.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: bazaar\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-09-07 08:08+0300\n"
+"POT-Creation-Date: 2025-09-08 02:13+0300\n"
 "PO-Revision-Date: 2025-08-12 22:05+0300\n"
 "Last-Translator: Automatically generated\n"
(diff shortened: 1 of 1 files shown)


kolunmi:
idk how to do it either

Saltane:
Oh, this is unfortunate. I think I may have some luck with `distrobox`, just need to pick a distro with correct libs and stuff.

The `.devcontainer` thing works, but I don't know how to spawn Bazaar window from it.

kolunmi:
@Saltane you can also just test with the flatpak!!

```
just build-flatpak
```
then it will install on your system, if you want to remove it after checking, do :
```
flatpak uninstall io.github.kolunmi.Bazaar
```

Hope this helps <3

kolunmi:
Oh also you spawn a window with 
```
bazaar window --auto-service
```

Saltane:
@kolunmi it does in fact look a lot more straightforward than whatever an awkward distrobox-dualwield I was cooking here. Thanks for the help.

No luck though, unfortunately. It seems to be building everything correctly, until it starts compiling Bazaar itself. It spends a few minutes on that and spits an error, stating «the file is too large» and also very long ugly path, as a treat. Unless there is an easy fix, will stick with releases from distro updates.

Speaking of which, Bazzite updated as I was fiddling with things. `0.4.8` looks cute, I like it.

PR #341, opened 2025-09-08
Category buttons rework

AlexanderVanhee:
This PR improves the usability of the category buttons.

It introduces a category tile, styled somewhat similarly to an app tile, which automatically adapts itself based on a `flathub-category`.

To support this, `flathub-category` now has additional properties for a human-readable name and an icon name. Since the Flathub API only provides the raw category name, these values are resolved through a lookup table. But it should still handle the case where the API response change or add a new category.

Some stylings and icons are taken from GNOME Software and then adapted for this use case.

image

image


kolunmi:
Astounding. Tysm

PR #342, opened 2025-09-08
fix(po): fix Ukrainian translation

Saltane:
A couple oopsies, one revision, and whatever `translators.sh` decided to do.


kolunmi:
:ukraine:

PR #343, opened 2025-09-08
Change image-missing-symbolic to better alternatives

AlexanderVanhee:
I changed out the symbolic placeholder icon for app icons to show nothing instead. This should hopefully make image loading less distracting, as it's just an image appearing compared to an image switching out. This follows the pattern most browsers handle it. It should also not cause any layout shifts.

I also changed out the symbolic icon of the update and install pages to be a fully colored alternative, making it a better fit with other entries on that page that do have colored icons.
Screenshot From 2025-09-08 14-49-51
image

### DIFF
--- a/src/bz-async-texture.c
+++ b/src/bz-async-texture.c
@@ -228,7 +228,7 @@ paintable_snapshot (GdkPaintable *paintable,
                     double        width,
                     double        height)
 {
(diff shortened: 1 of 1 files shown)


kolunmi:
I merged this manually but it seems gh didn't pick up on it :thinking: 

I'll close this but the code is in master now

PR #345, opened 2025-09-08
Update Brazilian Portuguese Translation

Lucosec:
Fixed some things. Everything is translated properly now.


kolunmi:
Thanks so much!

PR #348, opened 2025-09-09
chore(po): Update for italian translation

pbluez:
Update for version 0.4.8!

- Missing translations
- Bugfix

With this release, I left some probably dev strings untranslated (e.g. Bazaar Inspector, Active Curated-Configs, etc.) for a better experience and also because I'm not able to see where they are located.
Any feedback on this would be appreciated.


kolunmi:
Thank you!

PR #350, opened 2025-09-09
UI: Move categories from a dialog to a page

AlexanderVanhee:
his PR improves the usability of per-category catalog views by switching from dialog to page navigation. To achieve this, I implemented a switch from `Adw.ViewStack` to `Adw.NavigationView` for hierarchical navigation features. This change could also lead to generally better navigation in the future if some things are changed.

The page layout design is similar to that of the Flathub or curated pages.

Feel free to ask for changes.

image


kolunmi:
Thank you!

PR #355, opened 2025-09-11
translation: pt-BR update

victorsouzaleal:
image


kolunmi:
Interesting place for a face reveal but I applaud your courage

PR #356, opened 2025-09-11
Add install more button

AlexanderVanhee:
This PR adds a separate install button for situations where the user can both remove and install the application at the same time when multiple repos are available. This should help reduce user confusion.

image


kolunmi:
Amazing, thanks!

PR #361, opened 2025-09-12

Saltane:
This PR comes with no additional patents and license fees.


kolunmi:
thank goodness I was worried

Issue #364, opened 2025-09-13
po: strings remain untranslated


Saltane:
I noticed some releases have some entries untranslated. Initially, I thought it might be because entries in question were translated in-between update cycle and thus did not land to my machine just yet. I took my time to see if that is the case, and after the latest update, they remained untranslated:

<img width="513" height="317" alt="Image" src="https://github.com/user-attachments/assets/8bb20d9d-61f9-4270-be70-cd148c84c816" />
<img width="557" height="144" alt="Image" src="https://github.com/user-attachments/assets/764938f4-7a7e-430f-8e84-854c0fcd7696" />
<img width="223" height="89" alt="Image" src="https://github.com/user-attachments/assets/73079c4e-d2c5-4b39-a4ee-da3ad65e9560" />

I poked this issue, and noticed that only entries with `#, fuzzy` tag(?) in them were faulty, like this one:

```po
#: src/bz-preferences-dialog.blp:17
#, fuzzy
msgid "Git Forge Star Counts"
msgstr "Вподобань на сторінці розробки"
```

So I went ahead, cleared them out, and tested translation again. Now it all worked:

<img width="510" height="318" alt="Image" src="https://github.com/user-attachments/assets/e3bccaff-a553-4979-9f83-7dc065c68c90" />
<img width="559" height="133" alt="Image" src="https://github.com/user-attachments/assets/c860c208-555c-4f86-8866-d37a9a1680b2" />
<img width="237" height="86" alt="Image" src="https://github.com/user-attachments/assets/b7f0fd9c-c851-4b45-ab0d-4df9837fa4b2" />

I will be submitting fixes shortly soon, however…

### Issue

This issue seems to be originating from whatever `translators.sh` is doing when updating existing translation. It does not seem to be the case for freshly generated `.po` files. I don't know if updating manually behaves differently. Until fixed, translators must check updated lines for the `#, fuzzy` presence and remove it when applicable.

Saltane:
The issue seems to stem from how `gettext` suite merges changes with already present translation by default. That is, tries to guess which old strings may be valid translations for new lines and marks results with `#, fuzzy`.

One solution has been [sitting there on Stack Overflow](https://stackoverflow.com/questions/7372414/removing-all-fuzzy-entries-of-a-po-file) for a decade and a half. It involves using a `msgattrib` command from `gettext` suite…:

`msgattrib --clear-fuzzy --empty -o /path/to/output.po /path/to/input.po`

[...]

PR #365, opened 2025-09-13
Fix untranslated strings

Saltane:
This PR fixes some existing translations, affected by #364. 
It does not address the origin of the issue though.

:warning:  Tested with Ukrainian and Czech translations only.


kolunmi:
Thanks!

PR #377, opened 2025-09-20
(UI) Make share dialog use action rows

AlexanderVanhee:
This gives the dialog a more standard look and feel, and it also allows the use of activatable-widget property.
This moves the creation of the rows into code rather than a factory, since PreferenceGroups sadly can’t be created that way. 

image


kolunmi:
Thank you!!!

PR #378, opened 2025-09-20
(chore) Add correct icon css classes

AlexanderVanhee:
Adds the `icon-dropshadow` and `lowres-icon` classes to app icons for better visibility in light mode.

See:
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.3/style-classes.html#app-icons


kolunmi:
Thank you!

PR #380, opened 2025-09-21
(UI) Add toast on screenshot copy

AlexanderVanhee:
Adds a function in `BzWindow` for displaying toasts on the main window, and uses it to display a 'Copied!' toast on screenshot copy. Adding the function makes it so you don't need to scour the widget tree for the toast overlay every time you need to display a toast.


kolunmi:
Awesome! Thanks!

PR #381, opened 2025-09-21
(Refactor) Make use of properties of Adw.Toggle

AlexanderVanhee:
Make use of the label and icon-name properties of Adw.Toggle widgets instead of replicating them via a child widget. This also fixes the visual issue where the icons were not dimmed when the toggle was disabled.


kolunmi:
Looks awesome!!! Thanks

PR #384, opened 2025-09-22
po: Update French translation

HarmfulBreeze:


po/fr.po
```diff
@@ -659,26 +746,32 @@ msgstr "Minimiser la borne inférieure"
...
+#: src/bz-transaction-view.blp:173
+msgid "Updated Size"
+msgstr "Taille après mise à jour"
 
-#: src/bz-transaction-view.c:121
-#, c-format
-msgid "%s to install"
-msgstr "%s à installer"
+#: src/bz-transaction-view.blp:186 src/bz-window.blp:299
+msgid "Update"
```
HarmfulBreeze, line 763: There should be two different strings for "Update" as it is used both as a noun and as an infinitive.
The noun is translated into French as "Mise à jour" and the infinitive as "Mettre à jour".

I decided to fall back to the infinitive as it would otherwise be a bit awkward.

kolunmi:
Thanks a bunch!!

PR #401, opened 2025-09-28
(UX) Refine the Flathub views to be denser and have less partial lines

AlexanderVanhee:
This PR aims to make Flathub views less awkward by ensuring complete rows are always displayed and adding some other minor visual tweaks:

- Added a `max-lines-per-row` parameter to `dynamic-list-view` for better layout control.
- Displaying the app of the day together with apps of the week to ensure a divisible 6 entries.
- Updated category pages to fetch 60 entries instead of 50, so the total is divisible.
- Adjusted text weight and size in app tiles for improved readability and a cleaner look.
- Updated collections to only fetch 96 instead of a 100

### DIFF
--- a/src/bz-app-tile.blp
+++ b/src/bz-app-tile.blp
@@ -40,12 +40,13 @@ template $BzAppTile: Button {
         Label {
           css-name: "app-tile-title";
           styles [
-            "title-2",
+            "heading",
           ]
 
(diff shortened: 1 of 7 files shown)


kolunmi:
Looks amazing, thanks so much!!!

PR #402, opened 2025-09-28
fix installed page list outline ring

AlexanderVanhee:
The outline ring when navigating with the keyboard didn’t follow the card because of the extra padding added by the `ListItem` widget. This CSS fixes it.

I believe I tried to fix this before with [this](https://github.com/kolunmi/bazaar/commit/3a923410ab76198a6426d45bbb11023b936b84fd), but unsuccessfully.


kolunmi:
:+1:

PR #403, opened 2025-09-28
(UI) Improve visiblility of add-on feature

AlexanderVanhee:
People seem to sometimes miss the existence of the add-ons feature. This patch tries to address that by adding a button on the `full-view` page to open the add-ons dialog.

It also moves the add-ons button out of the popover dialog on the installed page. Normally, I would avoid having three buttons in a row, but since apps with add-ons are relatively rare, I think this is acceptable.

image

image


kolunmi:
Absolutely amazing, thank you so much!!

PR #404, opened 2025-09-28
Add installed icons to app tiles

AlexanderVanhee:
Since the search entry shows an icon when an app is installed, I made it so that app tiles on the Flathub page display them as well. I also replaced the checkmark with a design that was specifically made for installed apps.

image

image


kolunmi:
AWESOME thanks :)

PR #405, opened 2025-09-29
Calculate detailed app tiles text color based on luminance of background accent color

AlexanderVanhee:
This should hopefully ensure that the contrast between the text and background color is always maintained.

image


kolunmi:
awesome

PR #407, opened 2025-09-29
Rework Version History

AlexanderVanhee:
This PR redesigns the version history section of the BzFullView.  
The biggest actual change is replacing the list of issues with the release description, parsed in the same way as the long description.  

Release descriptions are included more often than issue lists and are also much more user-facing.  

I also switched to a boxed list design again, which fits better with the other design elements. Unfortunately, `boxed-list` does not work with `ListView`, so we have to construct it manually in code.

I don’t think we need to hide the version history behind a dialog, since it’s already at the bottom of the scrolled window and so it doesn’t make any other parts less accessible.

image

### DIFF
--- a/src/bz-entry.c
+++ b/src/bz-entry.c
@@ -949,11 +949,11 @@ bz_entry_real_serialize (BzSerializable  *serializable,
               const char *version                        = NULL;
               const char *description                    = NULL;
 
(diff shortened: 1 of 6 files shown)


kolunmi:
Amazing, once again

PR #408, opened 2025-09-29

Saltane:
YouTube is being very unserious player in supply chain for `ytdl`.


kolunmi:
fr whats their issue

PR #411, opened 2025-09-30

Saltane:
For Linux on desktops to succeed, coreutils, sudo, and curl need to be rewritten in vibecode. Things like that seem to be booming in popularity nowadays.


kolunmi:
I'm surprised no one has hopped on the opportunity to call a programming lang "vibecode" lmao

PR #416, opened 2025-10-01
rework graph UX

AlexanderVanhee:
This PR reworks the "Downloads over time" graph by changing some styling and formatting of dates. It also removes the top and bottom bounds settings and tries to find nice automatic min and max values instead.

https://github.com/user-attachments/assets/4e3d9968-4f65-44db-8c54-4fa48f9930f6


kolunmi:
This is very much an improvement, thanks!

PR #417, opened 2025-10-01
(UX) Improve date formatting in release history

AlexanderVanhee:
This should hopefully make life easier for people who can’t read ISO 8601 dates.

Showcase:
image


kolunmi:
nice nice, looks a lot more readable this way :-)

PR #420, opened 2025-10-02
Make "Show More" buttons navigate to separate page

AlexanderVanhee:
This PR changes the "Show More" buttons on the Flathub page to be positioned under their section. Clicking these buttons now opens a new page with the entries for that section. This is in line with how the Flathub site works.

To make the implementation more sensible, I changed the category page to be more generic, taking only a title and a list of apps.

https://github.com/user-attachments/assets/d04879c1-4f3c-4b39-9df0-b6b6f5bfa330


kolunmi:
lgtm

PR #425, opened 2025-10-03
chore: use yaml manifest from flathub 

renner0e:
I built a flatpak locally with some string changes to this source tree and that worked perfectly

image

EDIT: apparently gnome builder does not support yaml -_-


renner0e:
https://github.com/kolunmi/bazaar/pull/426

PR #426, opened 2025-10-03
 fix: use hardcoded config path

renner0e:
This is currently the only major difference between the manifests and would make it easier for me to test the flatpak that CI spits out

See:
https://github.com/flathub/io.github.kolunmi.Bazaar/commit/d46c650a51f466beded821f16e5cdc7a6b287cb0


kolunmi:
Awesome, thanks!

PR #431, opened 2025-10-04
Add Country Statistics to stats dialog

AlexanderVanhee:
This PR adds a new "Country Statistics" section to the statistics pop-up dialog. It works similarly to the one found on Flathub.

World map data, including translated region names, is loaded from JSON and drawn using GTK Snapshot, similarly to the other timeline graph.

Translated region names can be updated by manually running a Python script. This script is currently not run during the build process, as it only needs to be executed when a new language is added, and running it automatically would introduce an otherwise unnecessary build dependency.

image


kolunmi:
This looks AWESOME. Will inspect closer in a bit :eyes:

kolunmi:
Thanks so much! <3

PR #436, opened 2025-10-05
(UX) Improve workings and design of transaction dialog

AlexanderVanhee:
This PR fully reworks how the transaction dialog works. The dialog now only shows when needed. If only one option is available and the transaction is triggered via a button (be it app install or addon), the dialog will no longer appear.

It will now logically appear when either a transaction is triggered via search or multiple options are available.

The dialog itself has also been redesigned, mostly with the goal of creating a more user facing design. This includes hiding the radio buttons when only one option is available and reworking most text.

I also took the liberty of splitting the `ready_to_transact` into more maintainable pieces

image

image

### DIFF
--- a/src/bz-window.c
+++ b/src/bz-window.c
@@ -965,26 +965,28 @@ create_entry_radio_buttons (AdwAlertDialog *alert,
                             GListModel     *model,
                             gboolean        remove)
 {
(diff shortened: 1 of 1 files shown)


kolunmi:
awesome, so much better!!!

PR #437, opened 2025-10-05
(UI) Change remove button pulse effect color and comet size curve

AlexanderVanhee:
Changed the curve for the icon size of the comet to be inverse parabolic instead of linear, creating a nicer effect at the end of the animation. 

I also made the colour of the pulse a parameter of the overlay to be able to change the pulse colour of the remove button to match the button itself. Some accent colors previously caused an ugly clash with the red shade of the remove button.

https://github.com/user-attachments/assets/8f01cb3b-3d50-4b5d-b019-83999cf1778b

### DIFF
--- a/src/bz-comet-overlay.c
+++ b/src/bz-comet-overlay.c
@@ -37,6 +37,7 @@ struct _BzCometOverlay
 
   GHashTable *nodes;
   GArray     *pulses;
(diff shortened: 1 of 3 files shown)


kolunmi:
THIS LOOKS SO MUCH BETTER!!!

PR #442, opened 2025-10-06
Change preferences group to context bar

AlexanderVanhee:
This PR replaces most of the action rows seen under the app description with a context bar, a widget found in GNOME Software. This is done as these tiles are much better at using the given space than action rows, especially on wider window resolutions. These tiles also give us much more design freedom compared to the action rows, allowing for the addition of multiple symbolic icons, colours, etc.

As a GNOME app following the HIG, I find it sensible to copy some of the minor supporting design elements from other GNOME apps, at least until a better version is someday designed.

Feel free to ask for further changes.

image

image


kolunmi:
Looks great as always! Merging this for now :)

PR #443, opened 2025-10-07
(UX) Switch side panels to sidebars and use flat header bar designs.

AlexanderVanhee:
This PR tries to make the UI feel more modern by adapting a flat design for the header bars. To be able to practically achieve this in the current layout, a switch to sidebars is also necessary.

Feel free to ask for changes.

image

image


kolunmi:
Looks so cool! Thank you

PR #448, opened 2025-10-08
Update metainfo file to better follow flathub guidelines

AlexanderVanhee:
Changed the brand colors to have better contrast, and updated the screenshots to have a more reasonable window size, the current app design, and default decorations.

Screenshot From 2025-10-08 09-35-50

flathub
fullview
search


kolunmi:
Appreciated as always, thanks

PR #451, opened 2025-10-09
(UX) Polish up the context bar

AlexanderVanhee:
This PR changes some strings in the context bar to hopefully make the entries less confusing. The license tile now also opens the license page when pressed.


kolunmi:
Thanks! It appears that the spdx object doesn't actually serve any purpose, as it doesn't retain any state. Could you make its methods static functions instead? See bz-io.c, for instance

kolunmi:
Thanks for fixing! Looks good

PR #452, opened 2025-10-09
(UX) Add really useful app size dialog

AlexanderVanhee:
Screenshot From 2025-10-09 21-28-21


kolunmi:
Indeed, very useful :)

PR #455, opened 2025-10-10
feat: add nightly suffix to desktop file

renner0e:
adds this thingy here, makes it easier to differentiate between the version from CI and the regular one from flathub

image


kolunmi:
thx

PR #456, opened 2025-10-11

Lumaeris:
I love the comparison metric of random things :D


kolunmi:
Thanks, it is quite cute imo :D

PR #457, opened 2025-10-11

Saltane:
The bar to enter any software engineering job is high. Even interns need to know a bunch of stuff they would not realistically use outside a job. Or they need to be cool with drowning this world in AI. Brutal.


kolunmi:
:( yeah

PR #460, opened 2025-10-12
Fix keyboard not looping screenshot carousel

AlexanderVanhee:
It was focusing on the carousel widget, causing the logic overrides for the left and right arrow keys not to be run.


kolunmi:
lgtm

PR #463, opened 2025-10-13
(Bugfix) Minor screenshot dialog fixes

AlexanderVanhee:
Fixes the swipe gesture on trackpads to no longer weirdly zoom but swipe the carousel instead (pinch to zoom should still work). 

Also fixes the race condition issue where the first screenshot was shown instead of the selected one by building the carousel around the index given.


kolunmi:
Lgtm, thx

PR #466, opened 2025-10-14
(UI) Fix some description rendering things

AlexanderVanhee:
Made some minor visual adjustments, such as dimming the markers and aligning them to the top for multi line sections. Also switched to a solid circle for the bullet point.

Also, text is now selectable again, and the renderer is now also used for the release descriptions as well. Allowing us to fully remove the old parser code.

image


kolunmi:
Awesome, thanks

PR #470, opened 2025-10-15
Change corner radius of inner bar when in  pending state

AlexanderVanhee:
Screenshot From 2025-10-15 21-52-48
Screenshot From 2025-10-15 22-17-00


kolunmi:
thx!!

PR #474, opened 2025-10-16
Decrease button padding in the transaction sidebar

AlexanderVanhee:


kolunmi:
looks great, thanks!!

PR #479, opened 2025-10-17
Update Japanese text

camegone:
Add new text and change datetime format against broken month of dates. (I hope I would not break it again)
image


kolunmi:
Thank you!!!

PR #480, opened 2025-10-17
Use translator-credits

yakushabb:
Let's translators handle their information per translation.

- Easy to maintaining.
- More mainstream.
- Users only see language specific translator.

### DIFF
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -458,36 +458,11 @@ bz_application_about_action (GSimpleAction *action,
     /* This array MUST be NULL terminated */
     NULL
   };
-  const char *translators[] = {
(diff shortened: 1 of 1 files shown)


kolunmi:
Thanks for the PR!!! I'm going to hold off on this for now (I want to make sure all the translations are fixed for this first) but we should definitely switch in the future <3

kolunmi:
Btw, @yakushabb , by what name would you like to be credited with for now? Since you contributed Turkish translations I believe

ghost:
My name is Sabri Ünal

kolunmi:
Awesome, thanks! https://github.com/kolunmi/bazaar/commit/db893c5b9369fdf6a98d553c0fc12f7b01e24795

PR #482, opened 2025-10-17
Add a shortcut for preferences

yakushabb:
See also: https://developer.gnome.org/hig/reference/keyboard.html


kolunmi:
oops missed this, thanks!!

PR #485, opened 2025-10-18
(UX) Improve UI of transaction sidebar

AlexanderVanhee:
This PR tries to make the transaction sidebar nicer to look at by making margins and padding more consistent, hiding unnecessary elements like successful finished operations that are part of the app itself and the technical operation status.

This should hopefully make the sidebar less claustrophobic.

It also adds back the distinctions between the 3 transaction types in the UI.

Screenshot From 2025-10-18 16-45-36


kolunmi:
Thank you!

PR #486, opened 2025-10-18
fix: use full category name on more button on flathub view

renner0e:
I think this is a little awkward

image


AlexanderVanhee:
What if we did the opposite like it is on the Flathub site and changed it to "New", "More New", "Updated" and "More Updated".

AlexanderVanhee:
I will close this PR as this seems stale

PR #491, opened 2025-10-18
(UI) Change kind indicator from sub icon to coloured pill

AlexanderVanhee:
Having a little icon in the bottom right of the app icon looks weird to me. That's why I changed it to a pill to be more inline with other UI elements and GNOME UI patterns in general.

image

(Both are shown here for the showcase)


kolunmi:
Awesome!!

PR #492, opened 2025-10-19
Add Bulgarian translation

twlvnn:


kolunmi:
Thanks so much!! :bulgaria:

PR #493, opened 2025-10-19
Updated Bulgarian translation

twlvnn:
Apologies for the duplicate pull request, it turns out I used an outdated PO file and had to update the translation so its more recent.


kolunmi:
No worries!

PR #494, opened 2025-10-19
Rework fake shadow to only show at right times

### LINKED ISSUE #490: Shadows in header in App detail view
There is a shadow under header in app detail view. It would look better if it was removed.

Image

AlexanderVanhee:
This PR manages the fake shadow under the header bar better based on the scroll position of the 3 scrollbars on the full view page.

Fixes #490


kolunmi:
lgtm!!!

PR #496, opened 2025-10-20

Saltane:
Bazaar hasn't been updated for like three Bazzite updates now. Gotta dogfood myself using virtual machine.


kolunmi:
you can always install the flatpak from flathub!! `flatpak install io.github.kolunmi.Bazaar`

Lumaeris:
> Bazaar hasn't been updated for like three Bazzite updates now.

That's because Bazaar now relies on library Fedora 42 repos doesn't provide, so you should wait until Bazzite 43 comes out.

PR #508, opened 2025-10-22
Redesign the add-ons dialog to use a action rows

AlexanderVanhee:
This makes the dialog more inline with the share dialog and what is usually done with other Adwaita apps for listings like this. This sadly once again moves much of the layout out of the blueprint files due to ListView not supporting boxed lists.

image


kolunmi:
coolio, thanks

PR #510, opened 2025-10-22
(UX) Rework releases list

AlexanderVanhee:
This PR makes it that only the latest release is displayed on the app page, moving older releases to a separate history dialog. This change is for cases where apps would show an excessively long version history, which becomes problematic when we want to display "other apps by this author" beneath the version list.

Also this PR moves the creation logic out of the full view and into its own dedicated widget.

Screenshot From 2025-10-22 20-42-58
Screenshot From 2025-10-22 20-43-12

image

### DIFF
--- a/src/bz-releases-list.c
+++ b/src/bz-releases-list.c
@@ -164,24 +164,9 @@ create_release_row (const char *version,
 }
 
 static void
-bz_releases_dialog_dispose (GObject *object)
-{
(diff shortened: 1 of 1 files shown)


kolunmi:
Awesome, looks great!

kolunmi:
:hotdog:

KiKaraage:
"Show Previous Changelogs" might be more communicative for the button. (The "Version History" modal title can still be kept)

PR #512, opened 2025-10-23
(UX) Switch to one column in full view and place screenshots in a carousel

AlexanderVanhee:
This PR moves the screenshots into a carousel at the top of the page. Having them on the side often looked bad, as they were either too large or pushed to the bottom beneath all other content. Users who want a closer look at the screenshots are now better served by the built in screenshot viewer.

This change also gives us the freedom to implement better designs for this screen in the future.

image


snwh:
I pulled and tested this change seems good; there seems to be some cases where the size of the screenshots stretch the height of the carousel. It'd be good, IMO, to lock in a max-height for the carousel so it doesn't vary from app page to app page, gnome software seems to be 350px for example.

AlexanderVanhee:
> there seems to be some cases where the size of the screenshots stretch the height of the carousel

I swear i saw this issue during the development off this PR, but i can't seem to replicate it.

https://github.com/user-attachments/assets/c7f77d4c-2d35-48bd-a5da-0f149f9041a8

kolunmi:
I can't seem to replicate it either. I'll merge this for now and we can continue refine issues like these as separate prs.

kolunmi:
This looks awesome tho

PR #524, opened 2025-10-25
Strip trailing slashes from forge URLs when fetching stars

pojntfx:
This PR fixes a small bug I found where apps with trailing `/` characters in their forge URL would not have their stars displayed correctly (they would always be `0`). An example with [Sessions](https://flathub.org/en/apps/com.pojtinger.felicitas.Sessions):

Bazaar showing Sessions having 0 stars in the store page

After this PR, the issue is gone:

Bazaar showing Sessions having 6 stars in the store page

### DIFF
--- a/src/bz-full-view.c
+++ b/src/bz-full-view.c
@@ -771,15 +771,16 @@ debounce_timeout (BzFullView *self)
 static DexFuture *
 retrieve_star_string_fiber (GWeakRef *wr)
 {
-  g_autoptr (BzFullView) self    = NULL;
-  g_autoptr (GError) local_error = NULL;
-  g_autoptr (BzEntry) entry      = NULL;
-  const char      *forge_link    = NULL;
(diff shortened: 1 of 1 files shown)


kolunmi:
Hi, thanks for the pr!!

You are leaking some memory here; you assign dynamically allocated memory from `g_strndup` to `const char *`, then never free it. Could you fix this?

Also make sure not to commit the submodule thing

pojntfx:
Oh, sorry about that. Just removed the submodule. I missed `The caller of the function takes ownership of the data, and is responsible for freeing it.` from https://docs.gtk.org/glib/func.strndup.html - fixing that now.

pojntfx:
Alright, should hopefully be all fine now. Still getting used to GTK code in C, I usually do things in Go. I used `g_autofree` to free the string from `strndup`.

kolunmi:
lgtm! Thanks!!!

Issue #477, opened 2025-10-17
Try new approach to context tiles?


bertob:
Since this app is a bit more in flux than GNOME Software UI-wise, I was wondering if you'd be interested in trying a new approach to context tiles that we've been discussing on the design side.

Some of the background for this is [outlined in this issue](https://gitlab.gnome.org/Teams/Design/software-mockups/-/issues/22) but tl;dr it's a problem that some relatively important information (e.g. download size, safety rating, and device support) is currently not shown prominently enough on the app pages in both Software and on Flathub. It's currently kind of moot in Bazaar, but if/when the app page is redesigned to follow the HIG and moves away from odd multi-column layout, the same considerations will apply there.

If this approach ends up working out in Bazaar I assume we would eventually switch to it everywhere.

Some examples for how this approach would look with real apps (includes some speculative things that we don't have metadata for, such as a11y):

<img width="1096" height="447" alt="Image" src="https://github.com/user-attachments/assets/13a5a25f-6de0-4142-af06-375c9875929e" />

<img width="1096" height="447" alt="Image" src="https://github.com/user-attachments/assets/8dba8887-0a72-4eb1-bd37-ed0b07cad8c6" />

[Complete mockup](https://gitlab.gnome.org/Teams/Design/software-mockups/-/blob/master/context-tiles-refresh-tbernard.png?ref_type=heads)

Waytonn:
I do understand the mindset of risk in this instance, but for new or more casual users, they'd possibly be confused or uncertain about installing "high risk" apps

ramcq:
I love the aesthetic, I think my concern would be the same - how do people react to these cues if we make them more prominent. Could we put some mockups on e.g. usertesting.com and see how 5-10 people react? "Would you install this app?" etc

bertob:
> casual users, they'd possibly be confused or uncertain about installing "high risk" apps

That goes both ways, we've had a lot of feedback that the status quo is misleading/dangerous because if you don't scroll down and just hit install you never learn about the risks.

Obviously the usual disclaimers about how we're still missing portal APIs for some stuff, apps need to be updated to use them, etc. apply, but I think most of that is mitigated by limiting the higher risk levels to proprietary apps / EOL runtimes.

If someone wants to do research on this I'd be happy to provide A/B mockups :)

kolunmi:
This was implemented and will be available in the next release

PR #528, opened 2025-10-26
(UX) Rework the context tiles

AlexanderVanhee:
This PR implements context tiles in a similar way to the design seen in #477 by @bertob, though it deviates from the design in some areas. Some data isn't available yet in the app or AppStream in general (or at least that's what I believe), and in some cases our needs differ from GNOME Software's.

I've also been struggling about how the tiles should function as buttons. I'm not quite happy with the current state yet. I might try making it so the current button area remains the touch target, but only the pill actually reacts visually to hover/active states.

<p align="center">
  image
</p>

<p align="center">
  image
</p>

<p align="center">
  image
</p>

<p align="center">
  image
</p>

(Don't forget to squash)


kolunmi:
amazing :heart:

bertob:
I'm not sure if it's better to comment here, in the issue, or in a new issue, but I gave it a spin and here are some thoughts:

- First of all, this looks dope, thanks for working on it :heart:
- I agree about the hover state not being ideal, and the idea above of trying to highlight only the pill sounds interesting!
- The last two tiles would be a better fit further down on the page, but since we need 6 tiles for the layout to work I guess it's maybe fine to keep them for now? Alternatively we could also try to see if the layout works with just 4, but my feeling is it'd look too empty
  - If we do keep the Downloads/month tile I'd use the same font size for the "K" as for the number, it looks odd without a unit
  - I'd also drop the `/mo` part of the label, the abbreviation is not a good fit stylistically - The full string, e.g. "2.34 K Downloads/month" could be available as a tooltip, and inside the dialog maybe?
    
image

- I think it'd be good to have tooltips for all context tiles with a summary of the tile state, e.g. "Works on desktop, tablets, and phones", "Suitable for all ages", or "High risk permissions"
- It's a bit odd that not every tile is clickable/behaves in the same way, would be good to have a dialog for every tile
  - It'd be good to avoid linking to websites from the tiles, it's unexpected and a potential privacy issue
- The license tile is a bit odd with labels like "Free", which are probably misleading/confusing to most people, but I guess that'll go away anyway once it's replaced by a safety tile

image

- The symbolic icon for "All Ages" feels a bit odd to me, it draws attention to the tile rather than way from it (which is probably what we want since there's nothing to worry about). I'm not sure exactly what would be better, but maybe just `3+` since nobody uses a computer under that age anyway presumably, or maybe just the word "All"?

image

PR #537, opened 2025-10-28
Update de.po to fix .desktop Keyword entry

RurPet:
Added a ; at the end of the keyword entry in line 32.

This **might break** things because i'm unable to test.


kolunmi:
Awesome, thanks

PR #539, opened 2025-10-28
(UX) Context row fixes

AlexanderVanhee:
This PR tries to address some of the issues mentioned by @bertob in #528, including:
- Changing hover and active states to only affect the actual lozenge (but the touch target area remains the same).
- Adding tooltips to each context tile.
- Making the "K" in the downloads-per-month button the same size as the text.
- Changing the "All Ages" symbol to "3+".
- Adding a dialog for device support.
- Adding a dialog for the license tile.

Things that still need to be done are
- Adding an age rating dialog
- Adding a safety (risk factor) tile with a dialog explaining permissions
- Adding a high risk dialog
- Removing the last two tiles (the repo one will probably be the first one to go)

image

image

image

image

### DIFF
--- a/src/bz-full-view.blp
+++ b/src/bz-full-view.blp
@@ -148,7 +148,7 @@ template $BzFullView: Adw.Bin {
                                 ]
 
                                 has-tooltip: true;
-                                tooltip-text: _("This flatpak is verified by the original developers of the software.");
+                                tooltip-text: _("Verified by the original developers");
                                 icon-name: "verified-checkmark-symbolic";
                                 icon-size: normal;
                               };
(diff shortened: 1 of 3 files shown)


kolunmi:
Looks great, thanks

bertob:
Very nice improvements :raised_hands: 

On the Downloads tile, I'd drop the space between the number and the K, that's how other stores do it too.

I'd also still drop the `/mo` part of the Downloads label, it just looks weird. It's less informative to only have "3.2K Downloads" there and have "per month" only in the tooltip, but seeing as it's not very important and mostly a comparative metric I think that's a worthwhile tradeoff (it's also what other stores do, see below). Could be nice to explore adding that information to the dialog somehow though, maybe with some other numeric metrics, rather than just the chart/map.

Play store web listing for example:

image

PR #555, opened 2025-10-29
(UX) Flathub home page & category page rework

### LINKED ISSUE #535: Don't group "App of the Day" with weekly carousel
Currently the Flathub "App of the Day" is added to the weekly carousel. It's a convenient place to put it since the exact layout of the Flathub website can't be replicated here because we don't have an equivalent to the setup tile, but it's not quite the editorial intent from upstream, and I think we could do something much nicer.

On the Flathub home page we try to loosen up the boring category lists with more visual sections, such as the mobile and games sections. I think we could do something similar for app of the day here, e.g. by making it a carousel with a single page somewhere further down on the page, e.g. between Recently Updated and Recently Added.

Quick hacky mockup:

Image

### LINKED ISSUE #538: Move away from category tiles on the store homepage
Currently the home page uses coloured tiles to display a link to each category of applications available, these are poor for discovering applications within each category. To make an improvement here Bazaar can display each category as a section not unlike how it already presents Popular, etc. apps.

### Display app categories as groups

Since Bazaar performs a lot better than other app installation software, we can follow what other app stores do (including Flathub website itself) and display groups of apps for each category--showing the popular apps for that category--with a view more button to allow users to browse further into the category.

This presentation of app categories will give users a better sense of the sort of applications available in each category without them needing to dive deeper. A drawback here is that it introduces more scrolling on the homepage but I think on the whole the benefits to discoverability outweigh that, since scrolling to browse is less involved than clicking into and out of categories to see what is in there as part of the browsing experience of the home page.

### Group Trending, Popular, New and Updated into one section

This change will create several new categories on the home page, but Bazaar could take advantage of the libadwaita [InlineViewSwitcher](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.InlineViewSwitcher.html) to merge the four "Trending", "Popular", "Recently Added" and "Recently Updated" into one section to make them easy to access right at the top of the store front page and mirror what the Flathub homepage itself does.

With "Recently Added" and "Recently Updated" becoming "New" and "Updated" respectively.

***

The mockup below shows how the new categories would all appear on the homepage alongside some of the other proposed changes, and pre-existing layout.

Image

AlexanderVanhee:
This PR reworks the Flathub home page, trying to closely follow the design mock-up made by @snwh.

I didn’t change the title in the header bar on the category pages, since subpages don’t have their own header bar yet, which makes managing this a bit annoying. This should be easy to fix once each page has its own header bar.

I also took the liberty to add some text to the featured tile mentioning that it’s the “App of the Day.”

I decided to keep 12 entries per section, only switching to 6 on narrow displays. This keeps the layout clean in a 4 column view while preventing sections from becoming too long on mobile.

image

image

image

image

Closes #535
Closes #538

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -34,6 +34,7 @@ src/bz-entry-inspector.c
 src/bz-entry.c
 src/bz-env.c
 src/bz-error.c
+src/bz-featured-tile.blp
 src/bz-flathub-category.c
 src/bz-flathub-category-section.c
 src/bz-flathub-page.blp
--- a/src/bz-featured-carousel.c
+++ b/src/bz-featured-carousel.c
@@ -1,4 +1,4 @@
-/* bz-featured-tile.c
+/* bz-featured-carousel.c
  *
  * Copyright 2025 Alexander Vanhee
  *
@@ -19,8 +19,8 @@
  */
 
 #include "bz-featured-carousel.h"
-#include "bz-featured-tile.h"
 #include "bz-entry-group.h"
+#include "bz-featured-tile.h"
 
 #define FEATURED_ROTATE_TIME 5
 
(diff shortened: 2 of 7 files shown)


snwh:
This is working great :tada: aside from the header title thing, which you already noted. Feels like a much bigger improvement to the store home page experience.

kolunmi:
excellent

PR #559, opened 2025-10-29
(UX) Change App menu order & grouping

### LINKED ISSUE #558: Change App menu order & grouping
The app menu could do with a little shuffling (pretend the <3 is an emoji, Inkscape didn't like the emoji):

Image

 - Reorder the Bazaar-specific actions to give the Donate button the first-item advantage--an advantage that wasn't good for Quit since it is immediately focused meaning accidental quits would/could be likely.
 - Group the standard app actions:  "Preferences", "Keyboard Shortcuts" & "About" together after a separator while putting the Bazaar-specific actions together at the top.
 - rename "Quit Service" just to "Quit Bazaar" since "service" is ambiguous--"what service?"--and move it to the very end below another separator since it's not likely people will do it often.

AlexanderVanhee:
This PR does exactly what's suggested in #558.

Closes #558 

image

### DIFF
--- a/src/bz-window.blp
+++ b/src/bz-window.blp
@@ -483,20 +483,22 @@ template $BzWindow: Adw.ApplicationWindow {
 menu primary_menu {
   section {
     item {
-      label: _("_Quit Service");
-      action: "app.quit";
+      label: _("_Donate to Bazaar ❤️");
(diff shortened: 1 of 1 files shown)


snwh:
LGTM

PR #574, opened 2025-10-30
Make toggle group on Flathub page wider

### LINKED ISSUE #569: Wider buttons for the home page toggle group
There is a desire to have wider buttons for the toggle group on the Flathub page, like this:

Image

Which I did using  GtkInspector to add a style class and some padding:

```
.flathub-page-toggles toggle {
    padding: 0 72px;
}

window.narrow .flathub-page-toggles toggle {
    padding: 0 12px;
}
```

The problem is there isn't another breakpoint to accommodate the wideness, so the text will truncate:

Image

Would be nice to have a solution here so we can do both, otherwise we could find another solution for the width.

AlexanderVanhee:
To achieve this i added an extra breakpoint width 900px max width.

Make sure to turn off whitespace changes when viewing this PR.

Closes #569

image

### DIFF
--- a/src/bz-flathub-page.blp
+++ b/src/bz-flathub-page.blp
@@ -6,365 +6,386 @@ template $BzFlathubPage: Adw.Bin {
     width-request: 360;
     height-request: 450;
 
-    child: Adw.ViewStack stack {
-      enable-transitions: true;
-      transition-duration: 400;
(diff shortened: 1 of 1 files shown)


kolunmi:
Thanks a bunch :)

PR #575, opened 2025-10-31
(po):ja update ja translation

camegone:
I updated the translation. Additionally, I (hopefully) fixed keyboard navigation was not working.


kolunmi:
Thank you!!! :jp:

PR #576, opened 2025-10-31
Add "On the go" section and mobile category

AlexanderVanhee:
This PR adds the mobile category and the "On the go" section as seem on the Flathub site. 

As the app is now adaptive, i would think it makes sense to also have this. It also brings a visual break in the many category sections on the Flathub page.

image
Screenshot From 2025-10-31 12-42-04
Screenshot From 2025-10-31 12-46-06


kolunmi:
AWESOME

PR #579, opened 2025-10-31
(UX) Improve wording on preferences dialog

### LINKED ISSUE #548: Adjust labels of Search options
This is some proposed changes to search options to make it sound a bit less technical and be consistent with one another, especially the "Debounce" preference which unless explained a little better is a bit of a mystery

Image

> **Free Software Only**
> Hide proprietary applications from search results
> 
> **Flathub Results Only**
> Limit search results to applications only available on Flathub
> 
> **Delay Search Results**
> Improve results performance by debouncing search terms

### LINKED ISSUE #549: Make the Star Count option a bit less technical
The way the star count option is presented is fairly technical, it could do with being humanized a bit:

Image

Within the info popover is the better place to include a brief explanation about how this works (or may not work).

Image

If this feature is reasonably reliable, I'd also suggest having it on by default.

AlexanderVanhee:
It changes the wording like the issues suggest, but does not yet turn the star count option on by default.

Fixes #548
Fixes #549

Screenshot From 2025-10-31 16-11-06

### DIFF
--- a/src/bz-preferences-dialog.blp
+++ b/src/bz-preferences-dialog.blp
@@ -20,6 +20,25 @@ template $BzPreferencesDialog: Adw.PreferencesDialog {
     use-underline: true;
 
     Adw.PreferencesGroup {
+      title: _("Search");
+
+      Adw.SwitchRow search_only_foss_switch {
+        title: _("Free Software Only");
+        subtitle: _("Hide proprietary applications from search results");
+      }
+
+      Adw.SwitchRow search_only_flathub_switch {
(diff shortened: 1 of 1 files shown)


snwh:
LGTM, but after building I think left aligning the text in the popover might better, and the Application Details section could be moved below the Search section, since it's a bit more advanced of an option.

AlexanderVanhee:
Should be fixed now!

kolunmi:
cool!

PR #587, opened 2025-11-02
Add Hungarian translation

rdbende:


rdbende:
Some comments on the current strings:

- It's strange to me, that names are translatable. I don't think there would be any reason to translate people's names, nor is it common practice in GNOME applications.
- It would be good, if the "More Something" strings were separate strings, like on Flathub, and not "More %s", since:

    1. Not all languages use the Title Case, so for them, simply substituting the category name looks weird
    2. I need to use the plural form for the category title, but not for the "More Something" buttons. E.g.: the button label for "Segédprogramok" (Utilities) should be "Még több segédprogram" instead of "Még több Segédprogramok" in Hungarian.

- The "Installs:" label in the timeline views tooltip should be a formatted string to accommodate languages, where the numeric value would better be placed before the label.

kolunmi:
Thank you, I will make a note to fix these issues. I appreciate the work you did to make this translation! :hungary:

PR #594, opened 2025-11-03
Split "More <category>" strings to better accommodate different translation needs

rdbende:
As described in #587


AlexanderVanhee:
Should now work with the category rework i did yesterday.

kolunmi:
Awesome, thanks!

PR #595, opened 2025-11-03
Rework Categories

AlexanderVanhee:
This PR makes it so the Trending, Popular, Recently Updated and Recently Added collections get treated as proper categories, this makes it easier to reuse code that is already available for categories.

It also makes it feasible to add these to the search page.

Screenshot From 2025-11-03 21-45-34


kolunmi:
LGTM :heart:

Issue #593, opened 2025-11-03
Search Rework UI papercuts


snwh:
The Search rework is looking great nice work!  I have a few UI papercuts that could be fixed though:

**Button Styles**

<img width="365" height="151" alt="Image" src="https://github.com/user-attachments/assets/370788b3-b57a-44e8-b94a-0ab099ce0e98" />

Remove button doesn't need the destructive style as it has a confirmation dialog with the final remove action with the destructive button style; this could probably just use the remove/delete icon instead of the label also.

<img width="365" height="151" alt="Image" src="https://github.com/user-attachments/assets/f1277737-ffae-498d-9ed3-e4804f9fb220" />

It was mentioned in Matrix but we could probably drop the suggested style for buttons here, especially also has there is a confirmation dialog with the final action in suggested style.

**No Image/Empty State**

Some apps/results have no screenshots, which just show up as a grey box:

<img width="358" height="320" alt="Image" src="https://github.com/user-attachments/assets/c83666b4-6fbd-4f4c-af1e-6169dd4cb392" />

...could display a large (96px) symbolic icon (such as  `image-missing-symbolic` which is a standard icon) to fill that space.

<img width="408" height="338" alt="Image" src="https://github.com/user-attachments/assets/085088a5-ab68-477b-a7e9-2845f536edbc" />

**Focus State on Initial Card Button**

Not sure how easy this would be to do, but seeing as pressing enter activates the button on the first card it might be worthwhile trying to have a focus style/ring present on the button in that first card.

**No Section Header**

<img width="1018" height="110" alt="Image" src="https://github.com/user-attachments/assets/aeee30ff-0110-42e8-b03b-450077edb78f" />

The designs in #578 had a **Browse Categories** header for the tiles section, this was there to cue users that clicking on these will take them elsewhere to browse a section--and not imply these are categories to search.

**No Results Count**

The designs in #578 also had a label above the grid showing the result count for a given term, I think it would be good to have that eventually.

AlexanderVanhee:
I have split the remaining problems into their own subissues to make them more manageable.

PR #598, opened 2025-11-04
Minor search fixes

AlexanderVanhee:
Made the changes like asked, does not include the search result amount as `GridView` doesn't seemingly support headers.

image

### DIFF
--- a/src/bz-rich-app-tile.blp
+++ b/src/bz-rich-app-tile.blp
@@ -109,7 +109,6 @@ template $BzRichAppTile: Adw.Bin {
           visible: bind $invert_boolean($is_zero(template.group as <$BzEntryGroup>.removable) as <bool>) as <bool>;
 
           styles [
-            "destructive-action",
             "circular"
           ]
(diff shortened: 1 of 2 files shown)


rdbende:
> Made the changes like asked

This again doesn't quite follow the mockups, does it? Or was this discussed in somewhere else than #593?

> as GridView doesn't seemingly support headers.

Couldn't it just be any label outside of the `GridView`, tho?

AlexanderVanhee:
The buttons were further discussed in the Matrix, with this design being my interpretation of the discussions conclusion.

The label could of course be placed outside of the `GridView`, but it would be non-scrollable and I think that would look weird. I also tried placing the result amount in the search bar and it didn't work much better either.

rdbende:
Ahh, right. Because the `GridView` needs to be the sole child of a scrollable to work.

kolunmi:
@rdbende email me

rdbende:
Sent

PR #599, opened 2025-11-05

Saltane:
Lokalize is a bit janky at working with long strings, but an improvement in workflow otherwise.


kolunmi:
Sorry the strings are changing so much :sob:

Saltane:
Don't worry about it. If things need to be changed, let the change be, no matter how often. The toolset I use now can handle it without any inconvenience.

Issue #551, opened 2025-10-29
Some UX refinements for the App Page header


snwh:
Nice to see the changes to the application listing page rapidly land, and with the new context tiles, but I've got a few things to comment on about them and the overall header layout.

**Move "Support" to a context Tile**

The pill presentation underneath the application name/author doesn't invite you to click it, and is a bit of a misuse of that pattern, moving it to a tile makes it more prominent.

I'd also throw another icon in there alongside the heart to clarify it's a donation button in most cases

**Drop the Flathub Tile**

This link already below in the applications links section so it's not particularly necessary to display as a context tile here (it's also consistent across nearly every application listing so it doesn't make it stand out as a unique property).

**Label Changes**

`Download -> Size`
The former sounds like an action rather than a quality of the application, so changing it to "Size" I think would be appropriate

`Downloads/mo -> Downloads`
Moving the "/mo" into the pill itself cleans up the presentation of the tile, IMO.

**Age Rating**

The "All Ages" case is different from other ratings. Is there a reason it's showing the "E" or "EC", or whatnot label while keeping "Age Rating" in the sublabel?

<img width="115" height="90" alt="Image" src="https://github.com/user-attachments/assets/1ebe5261-0320-4ee2-a031-c3cb817af97b" />

***

<img width="1226" height="390" alt="Image" src="https://github.com/user-attachments/assets/50199c86-dc82-4e60-a929-41c5ae905082" />

PR #602, opened 2025-11-05
(UX) Add "Other Apps by Developer" section and minor context tile fixes

AlexanderVanhee:
This PR adds a section that displays other apps by the same developer.

Screenshot From 2025-11-05 22-00-21

For the context tiles, I removed the space before “K,” added styles for age ratings starting from 12+, and changed “Download” to “Size,” as requested in #551.

Screenshot From 2025-11-05 21-14-00

### DIFF
--- a/src/bz-full-view.c
+++ b/src/bz-full-view.c
@@ -456,41 +456,51 @@ format_other_apps_label (gpointer object, const char *developer)
     return g_strdup_printf (_("Other Apps by %s"), developer);
 }
 
-static GList *
-filter_own_app_id (BzEntry *entry, GList *app_ids)
+static gpointer
+filter_own_app_id (BzEntry *entry, GtkStringList *app_ids)
 {
     const char *own_id;
-    GList *filtered = NULL;
+    g_autoptr (GtkStringList) filtered = NULL;
+    guint       n_items  = 0;
 
-    if (!app_ids || !entry)
+    if (!BZ_IS_ENTRY(entry) || !GTK_IS_STRING_LIST(app_ids))
         return NULL;
 
     own_id = bz_entry_get_id (entry);
     if (!own_id)
         return NULL;
 
-    for (GList *l = app_ids; l; l = l->next)
-        if (g_strcmp0 (l->data, own_id) != 0)
-            filtered = g_list_prepend (filtered, l->data);
+    filtered = gtk_string_list_new (NULL);
+    n_items  = g_list_model_get_n_items (G_LIST_MODEL (app_ids));
 
-    return g_list_reverse (filtered);
+    for (guint i = 0; i < n_items; i++)
+      {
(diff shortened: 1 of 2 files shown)


src/bz-full-view.c
```diff
@@ -281,11 +284,11 @@ format_recent_downloads (gpointer object,
     return g_strdup (_ ("---"));
 
   if (value >= 1000000)
-    /* Translators: M is the suffix for millions, \xC2\xA0 is a non-breaking space */
-    return g_strdup_printf (_ ("%.2f\xC2\xA0M"), value / 1000000.0);
+    /* Translators: M is the suffix for millions */
+    return g_strdup_printf (_ ("%.2fM"), value / 1000000.0);
```
rdbende:
```suggestion
    return g_strdup_printf (_ ("%.2fM/mo"), value / 1000000.0);
```

src/bz-full-view.c
```diff
@@ -281,11 +284,11 @@ format_recent_downloads (gpointer object,
...
   else if (value >= 1000)
-    /* Translators: K is the suffix for thousands, \xC2\xA0 is a non-breaking space */
-    return g_strdup_printf (_ ("%.2f\xC2\xA0K"), value / 1000.0);
+    /* Translators: K is the suffix for thousands*/
+    return g_strdup_printf (_ ("%.2fK"), value / 1000.0);
```
rdbende:
```suggestion
    return g_strdup_printf (_ ("%.2fK/mo"), value / 1000.0);
```

rdbende (commented):
The "unit" of monthly downloads is downloads/mo, so the /mo should be part of the pill label.

rdbende:
Hmm, actually, it would be nice to have the "K/MO" label with a smaller font size, same as with "MB" for the download size.

AlexanderVanhee:
> Hmm, actually, it would be nice to have the "K/MO" label with a smaller font size, same as with "MB" for the download size.

The "K/MO" thing really can't be done due to localisation and that  part was once small, but I was asked to remove that.

rdbende:
Hmm, what was the issue with localisation here?

AlexanderVanhee:
Not every language may have an abbreviation for "monthly downloads". This is fine for the subtitle, as there isn’t really a space limit, but you definitely do for the pill.

rdbende:
Well, in that case, as a translator I'd just leave the suffix blank, and add translate the subtitle as "Monthly Downloads", or smth like that.

kolunmi:
@AlexanderVanhee  Looks awesome, but I don't want BzApplication to have methods for structural reasons. The BzStateInfo object is there to hold the mapping factory. I can make this change if you want

AlexanderVanhee:
Should be fixed now @kolunmi

PR #606, opened 2025-11-06
Update category icons, change style and layout of tile view

snwh:
This PR replaces/updates some icons and changes the layout of the category tiles and makes the gradients consistent direction.

image


snwh:
image

Something in this PR is that play has a space in it or something, not sure where it's coming from.

kolunmi:
Looks amazing, thanks!!!

PR #609, opened 2025-11-08
Change screenshot dialog to page

AlexanderVanhee:
This PR moves the former screenshot dialog to a dedicated page. This improves the initial state, as the screenshot is now displayed larger instead of being cramped inside a dialog.

It also fixes an issue on mobile where the dialog sheet would sometimes close when users tried to use gestures to move the image, as some of these gestures overlapped.

image


kolunmi:
LGTM

PR #610, opened 2025-11-08

Saltane:
gettext's and Lokalize's formatting styles differ slightly, adds a few extra changes to the diff.


kolunmi:
:)) thanks

PR #613, opened 2025-11-08
(UX) Featured Banner: show screenshot at smaller sizes

### LINKED ISSUE #476: Banner: Show screenshot at smaller sizes
The new banner is really cool, but unfortunately the screenshots are not shown unless the window is huge, leading to awkward inbetween states like this:

Image

This isn't as much of an issue on the Flathub website since the browser is usually maximized, but in a native app it's a bit odd. It'd be good to ensure it's visible for as small a size as possible, at least down to about 900px or so.

AlexanderVanhee:
This PR allows the top halves of screenshots to appear at smaller window widths (around ~950 px). I already had a fix in the pipeline for this, but it previously caused layout warnings, which now seem to have disappeared, likely due to the recent navigation rework.

https://github.com/user-attachments/assets/f80128d6-c15f-4cce-820d-cd45031d1ef6

Fixes #476

### DIFF
--- a/src/bz-featured-tile.blp
+++ b/src/bz-featured-tile.blp
@@ -21,7 +21,7 @@ template $BzFeaturedTile: Button {
     Box content_box {
       halign: center;
       orientation: horizontal;
-      margin-start: 50;
+      margin-start: 10;
       margin-end: 10;
 
       Box {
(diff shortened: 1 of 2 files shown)


kolunmi:
cool thankssss

PR #615, opened 2025-11-08
chore(po): remove obsolete strings for russian locale

Lumaeris:
Mostly they wouldn't be used in any way anymore, plus some strings were translated based on context wherever they were used, not always literally meaning what it supposed to when reused once again.


kolunmi:
thanksies

PR #619, opened 2025-11-09

Saltane:
I frogor to split PR, but I think it's fine.


Saltane:
Removed commit in favour of #621

kolunmi:
thx again <3

PR #620, opened 2025-11-09
Use mask for the fading clamp

AlexanderVanhee:
Instead of drawing the gradient using the same color as the background to create a mask like effect, we now simply use an actual mask. This allows us to use this widget in different places with different background colors without issue.


kolunmi:
Thanks!!

PR #622, opened 2025-11-09
Installed page rework

AlexanderVanhee:
This is the first part of the App Library rework, it changes the installed tiles to be inline with the design.

Screenshot From 2025-11-10 11-41-21


kolunmi:
thanks, LGTM

PR #623, opened 2025-11-10
flatpak: Fix tags in the manifest

alatiera:
The type of the tags in the manifest in the schema is an array.

https://github.com/flatpak/flatpak-builder/blob/main/data/flatpak-manifest.schema.json


kolunmi:
Thanks!!!

PR #624, opened 2025-11-10
Handle (partial) failures of the Flathub API better

AlexanderVanhee:
This PR handles cases where the Flathub API fails for some reason, but the user is still online.
If only some requests fail, the affected widgets should now be properly hidden/disabled.
If all requests fail, special empty states are shown on the Flathub and search pages informing the user what they are still able to do.

Screenshot From 2025-11-10 13-44-02

Screenshot From 2025-11-10 14-18-23


kolunmi:
Perfect, thanks

PR #633, opened 2025-11-12
(Bug) Fix global progress bar colors

AlexanderVanhee:
Made it so the accent color option for the global progress bar now follows CSS. This fixes the issue where the accent color didn’t turn purple on Flathub styled pages.

I also fixed an issue where a global progress bar that has a child prevented the flags from showing up, falling back to just the accent color.


kolunmi:
<3

PR #635, opened 2025-11-12
Add EOL notices

AlexanderVanhee:
This PR adds EOL notices to the installed app tile and app page, similar to how GNOME Software handles them. The wording has been slightly adjusted since this concerns runtimes no longer receiving updates, rather than the app itself.

I also added a similar notice to the rich search tiles, as apps can be installed or removed directly from there.

Screenshot From 2025-11-12 18-09-22
Screenshot From 2025-11-12 18-09-04
Screenshot From 2025-11-12 18-08-44


kolunmi:
<3

PR #636, opened 2025-11-12
(Bug) handle apps without stats better

AlexanderVanhee:
This PR properly disables the stats button when no JSON data is found, like with brand new apps. Previously you were able to open an empty dialog.

image


kolunmi:
<3

PR #648, opened 2025-11-13
Port Age Rating Dialog

AlexanderVanhee:
This PR adds the age rating dialog, adapted from the one found in GNOME Software. Most description strings are localized by the Appstream library here, we only have to handle the "No Information" ones.

image
image


kolunmi:
Some style issues (don't prefix things with `bz` if they are local to the c file), but not a big deal. Looks good! Thanks!!

PR #652, opened 2025-11-15

Lumaeris:
[Tyrone would just like to interject for a moment [q8Qftb2O--Y].webm](https://github.com/user-attachments/assets/64c4e485-d6b9-41ba-9642-58634d87ac1a)


kolunmi:
This is true and factual

PR #658, opened 2025-11-16
Add tags to App Detail Pages

AlexanderVanhee:
Adds a list of tags (known as keywords in AppStream) to the full view for apps that provide them. Clicking on a tag makes an async request to Flathub to fetch a pseudo category containing all apps with the same tag.

image

image


kolunmi:
thanks!

PR #659, opened 2025-11-16
fix(po): fix Ukrainian translation

Saltane:
Stylistic revisions, also fixes some regressions introduced in #599 (I performed a lil shoot in the foot sorry)


kolunmi:
no worries!

Issue #670, opened 2025-11-19
Homepage papercuts


bertob:
The homepage is really coming together, thanks everyone for getting it this far 🙌 

A few minor papercuts I came across:

### Screenshot centering on the banner

I'm not totally sure why it happens (I assume it depends on the screenshot image size?), but sometimes screenshots have a bunch of whitespace to their right, whereas other times they go all the way to the edge. It'd be good to adjust the alignment to be more harmonious for different screenshot sizes, especially at the narrower end of the window size spectrum (it's less noticeable at large window sizes).

<img width="982" height="320" alt="Image" src="https://github.com/user-attachments/assets/b64fecb5-9506-4638-a4bf-b642d6ff2bfe" />

<img width="982" height="320" alt="Image" src="https://github.com/user-attachments/assets/bd2c8afd-ab86-4a0c-87dd-354af40ef717" />

### Switcher size

The trending/new/etc. switcher feels a bit too narrow in the layout, quickly tried just adding 3px more vertical padding and it helps a lot i think:

<img width="1510" height="794" alt="Image" src="https://github.com/user-attachments/assets/68a8a42b-697d-43e1-8086-b84d90dcb3ae" />

I'd also reduce the spacing between swichter and the grid below by a few pixels to make it clearer that they're part of the same section.

### More spacing between sections

The spacing between the "More $CATEGORY" buttons and the next titles is a bit tight, I'd increase it by 12 or 18px.

<img width="725" height="600" alt="Image" src="https://github.com/user-attachments/assets/5af03946-84f1-44c2-8ce8-384026340b07" />

### Tighter spacing between cards

The spacing between cards is a bit large on the mobile breakpoint. Could we reduce it to about half for that breakpoint?

<img width="410" height="600" alt="Image" src="https://github.com/user-attachments/assets/88fadea8-2f1a-4713-8427-8b1aea3fb74f" />

### Odd hover states on "Mobile" section

The opaque app card hover state on the "Mobile" section doesn't look great, I'd either make it semi-transparent like the Flathub website does or just use semi-transparent text color, like we do for hover states elsewhere.

<img width="1084" height="399" alt="Image" src="https://github.com/user-attachments/assets/48d9eccd-802b-495c-97f4-08878185a723" />

<img width="490" height="191" alt="Image" src="https://github.com/user-attachments/assets/c94101cb-5317-4211-ad9e-4fbc14e01cc5" />

### Tweak spacing around mobile illustration

[...]

PR #673, opened 2025-11-19
Fix some Flathub page papercuts

AlexanderVanhee:
This PR fixes the
- Switcher size
- More spacing between sections
- Tighter spacing between cards on mobile
- Odd hover states on "Mobile" section (in light mode)
- Tweak spacing around mobile illustration
- Mobile section doesn't use title case
- More spacing at the bottom of the page

points of issue #670.

To achieve tighter spacing between cards on mobile, I had to refactor how the dynamic list view works by shifting the spacing responsibility from its child widgets (AppTiles for example) to the `FlowBox`.

image

image


kolunmi:
thanks, wonderful as always

PR #674, opened 2025-11-20
feat(translators): add translator note regarding `fuzzy` marks

Saltane:
Solves #364.

Documents `gettext`'s behaviour that marks entries as fuzzy in a form of translator's notice.


kolunmi:
Thanks!

PR #677, opened 2025-11-20
(bug) Fix installed tile fallback icon margins

AlexanderVanhee:
The fallback icon had different margin sizes, causing installed tile rows without an app icon to be more vertically challenged than the others.


kolunmi:
Thanks!!

PR #696, opened 2025-11-21
Update French translation

theMimolet:
Bonjour !

I've added the translations for the French language for all the new content ! 
Including all the new buttons and options, new search page, all the menus and even did the ones for the age rating menu (that doesn't exist yet ?)

I've made people I know check it to be sure it's the best translation ever ^^

I'll try to keep up with the updates as soon as possible when they get out,

Lots of love from France,
Mimolet

PS : I've shared some of my WIPs in a discord server, they enjoyed it, as well as the overall design of the app that is just pretty af


kolunmi:
Thank you so much!!! <3 🇫🇷

PR #699, opened 2025-11-23
Hide the header bar task label 

AlexanderVanhee:
Like on the loading screen, I believe it’s best to hide this label as well. It also caused issues on mobile-sized windows

:shipit:


kolunmi:
hmmmm alright

PR #704, opened 2025-11-24
(UX) Add explanation popover for app verification

AlexanderVanhee:
This PR adds a popover next to the developers name on the full view explaining unverified apps and the different verification methods an app may have.

It also moves the fetching of verification status from the API to just getting it from the apps AppStream (at least for the ones coming from Flathub). Making the status appear instantly and allowing us to show verification on other pages quite easily.

image
image
image
image


kolunmi:
Awesome, thanks <3

PR #706, opened 2025-11-25
Add Cisgender and Straight flag

Aleksanaa:
Per the request of @BryanLunduke in https://x.com/LundukeJournal/status/1992803424090325343:

> If you’re thinking, “but… I’m straight… what flag do I pick?”  You don’t.  No special progress bar theme for you.

Yeah, this is a trivial issue. Let's fix it.

Note: I'm not very good at typography, and maybe we can find better parameters for UI, but at least it's fit now.

### DIFF
--- a/src/bz-preferences-dialog.blp
+++ b/src/bz-preferences-dialog.blp
@@ -11,7 +11,7 @@ template $BzPreferencesDialog: Adw.PreferencesDialog {
     condition ("max-width: 625sp")
 
     setters {
-      flag_buttons_box.max-children-per-line: 6;
+      flag_buttons_box.max-children-per-line: 7;
     }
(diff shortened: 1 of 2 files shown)


kolunmi:
I normally wouldn't respond, but you seem nice enough.

Thanks for the pr. I'm curious to know if you actually tested this, as you didn't actually implement the snapshot drawing logic. So it won't work.

Regardless, the individual you linked is a known far-right troll who wants engagement, and as such his opinion will not considered. He and his community will survive this.

If anyone is reading this and feels upset or left out by the existing selection of optional pride flags, I can definitely look into adding some straight ally flags in the future. Definitely doable.

I hope you have a wonderful day!

PR #707, opened 2025-11-25
Add Verified icons to (some) app tiles

### LINKED ISSUE #686: (UX) "Installed" checkmark can overlap app names
Part of #670

Should probably ellipsize names?

Image

AlexanderVanhee:
As we are primarily an app store for Flathub, I think it makes sense to add verified badges to app tiles, similar to how they do on their site, to promote upstream apps instead of community packaging efforts.

Also fixes #686 

image
image


kolunmi:
awesome, thanks again <3

PR #715, opened 2025-11-25
Add "We ♥ Games" section

AlexanderVanhee:
This PR adds a basic version of the “We ♥ Games” section found on the Flathub home page, without the other categories. I have not yet added those due to some UI and data blockers.

It also refactors how category app pages are created.

image
image


kolunmi:
amazing, thank you!!!!

PR #718, opened 2025-11-26
Remove title char limit on the app tile

AlexanderVanhee:
I thought this limit would fix the inconsistency column thing, but it just made the title look weird on some occasions.


kolunmi:
thanks!

PR #723, opened 2025-11-27
chore(i18n-de): update translation

renner0e:
still missing 10% but whatever, merge this whenever you happen to make a release


kolunmi:
Thank you!!

PR #729, opened 2025-11-27
chore: updated the dockerfile Ubuntu image and dependencies

theMimolet:
Hi ! 

I updated the Dockerfile to work with Bazaar's current dependencies !

So you know, it's able to create the container, run the meson command, but it fails building with ninja - and I'm not sure if I'm the one making a mistake or if Bazaar is currently in heavy development  :sweat_smile: 

But here you go, have a nice day !


kolunmi:
oh wow thanks!! I just use the flatpak manifest so this slipped under the radar :)

PR #734, opened 2025-11-28
(UI) Remove "view" style from curated page

AlexanderVanhee:
If I understand correctly, the `view` style was added to this whole page to mask the inherited `view` style of some elements. This worked well enough in older versions of the app, but since we moved to the flat header bar, the background colour change has become really distracting. So I used some CSS to remove the darker background entirely.

image


kolunmi:
Looks good! Thanks

PR #736, opened 2025-11-29
(Bug) Sort download statistics by date

AlexanderVanhee:
Flathub seems to have started sending statistics data without sorting it by date anymore. This appears to be unintended, since the Flathub site also doesn’t seem to handle it properly. Either way, it’s probably better to sort it by date on our end.

### DIFF
--- a/src/bz-entry.c
+++ b/src/bz-entry.c
@@ -2237,10 +2237,13 @@ query_flathub (BzEntry *self,
       g_steal_pointer (&future));
 }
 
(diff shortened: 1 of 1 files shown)


kolunmi:
Hi thanks!! Could you reformat the sorting function to fit with style conventions? Also Bazaar tends to cast the function pointer instead of the arguments inside the function

AlexanderVanhee:
Sorting function should be fixed now.

kolunmi:
Thank you!!

PR #742, opened 2025-11-30
Add various additional pride flags to Bazaar

Synby:
This PR adds the aroace, agender, genderqueer, intersex, demigender, and biromantic flags as optional themes for the global progress bar!

The demigender flag, while not official in the same way the the demiboy and demigirl flags are, is widely agreed upon as an umbrella flag and was made by the same person who made the gender-specific variants.

The biromantic flag is an alternate variant of the flag, as the more widely accepted version would look effectively identical to the bisexual flag if converted to a gradient as done here.

I wanted to also add the demisexual flag as well, but there isn't a good way to make it into a gradient without it looking identical to the asexual flag other than removing the black triangle entirely and I don't know whether or not would be a good idea so I left it out from this PR.

### DIFF
--- a/src/bz-global-progress.c
+++ b/src/bz-global-progress.c
@@ -621,13 +621,13 @@ bz_global_progress_snapshot (GtkWidget   *widget,
             6.0 / 7.0,
           };
           const float sizes[G_N_ELEMENTS (colors)] = {
-            0.0 / 7.0,
             1.0 / 7.0,
-            2.0 / 7.0,
-            3.0 / 7.0,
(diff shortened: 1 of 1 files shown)


kolunmi:
Hi!! Thanks so much for this pr, I'll take a look in a bit

kolunmi:
Fixed a few things!! Thank you for being polite

Synby:
oooh i missed that  heh, it worked despite it for some reason but thanks for fixking!

PR #749, opened 2025-12-01

theMimolet:
Hello ! 

Added the translation for the new pride flags + a few corrections and polish :+1: 

I just have one question, just to be sure : 
There is _Sync_ in English, but I'm not sure if this is for something static like a title or interactive like a button - right now I have it translated for a button.


Saltane:
It is a tooltip for header bar button, according to the source code. I presume it is nearby the update button on the top left, when there are updates available.

It would be better to have more contexts left by devs themselves.

theMimolet:
Thank you, it was the context I needed ^^"
And my translation still works so I'll mark it as ready :P

kolunmi:
awesome, thanks so much !!!

PR #755, opened 2025-12-03
(UX) Move manual sync to banner with explanation

AlexanderVanhee:
This PR replaces the icon button shown when on a metered connection with a banner that clearly explains what being on a metered connection means and what action they can take.

image


kolunmi:
thanksies :}

PR #758, opened 2025-12-03
(UX) Add a subtle colored tint to the carousel

AlexanderVanhee:
This PR adds a subtle tint to the screenshot carousel based on brand colors provided by the apps metainfo. The mixing values are defined as constants, so they should be easily tweakable.

image

image

image

image

image


kolunmi:
looks awesome as always!!

PR #762, opened 2025-12-04
(UX) Add captions to screenshot page

AlexanderVanhee:
This PR adds screenshot captions to the screenshot page. I have decided not to add them to the full view to spare the vertical space.

I also had to make it a separate list next to the list of screenshot paintables to avoid breaking the caching.

Screenshot From 2025-12-04 21-52-47
image


kolunmi:
amazing!

PR #764, opened 2025-12-05
Update pt_BR translation

koitorin:
Updating the Brazilian Portuguese translation to the new version of
Bazaar.

New flags! :3


kolunmi:
Thanks!!! :brazil:

PR #766, opened 2025-12-05
(UX) Few UX refinements for presenting Updates

### LINKED ISSUE #546: Few UX refinements for presenting Updates
### Pill to Show Update Count

I think it would be good to display the count for the number of updates available on the button once Bazaar has finished checking for updates

Image

### Spinner with Tooltip

Instead of displaying the spinner with the string inline on the header position it where the button would be and use a tooltip

Image

### Manual Refresh

Keeping a "Check for Updates" button always available in the header I think is a good idea, especially if/when automatic update checking is disabled.

Image

AlexanderVanhee:
I had to use a seperate box for the outline as css border created some weird aliasing artefacts.

Screenshot From 2025-12-05 20-28-07

Closes #546

### DIFF
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -1366,7 +1366,7 @@ respond_to_flatpak_fiber (RespondToFlatpakData *data)
         }
       else
         {
-          bz_state_info_set_background_task_label (self->state, _ ("Checking for updates..."));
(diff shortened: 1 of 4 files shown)


kolunmi:
:fire:

PR #769, opened 2025-12-06
(UX) Improve the user data removal experience.

### LINKED ISSUE #693: Allow for the deletion of user data.
Currently, users cannot remove user data without external means. I would say this is currently the biggest regression compared to GNOME Software.

This is what the uninstall dialog should look like (from #544):

Image

Something like this would also make sense on the Full view page:

Image

Having a leftover data subpage on the Installed page also seems really useful to me. Otherwise, the user is expected to hunt for all leftover data.

Image

AlexanderVanhee:
This PR improves the UI of the removal dialog and adds a banner below the screenshot carousel to inform the user that data remains for that app id, and to give them the option to remove it.

We still have to make sure this data gets moved to trash instead a full delete to protect user data.

I still believe we will need a way to view all apps with leftover data somewhere, as there is 0 way to list them otherwise.

Screenshot From 2025-12-06 18-00-16
Screenshot From 2025-12-06 14-33-52
Screenshot From 2025-12-06 14-33-34

Closes #693

### DIFF
--- a/src/bz-entry-group.c
+++ b/src/bz-entry-group.c
@@ -62,12 +62,13 @@ struct _BzEntryGroup
   int updatable_available;
   int removable_available;
 
-  gboolean   has_user_data;
+  gboolean has_user_data;
+
   DexFuture *user_data_exists_future;
+  DexFuture *reap_user_data_future;
 
   GWeakRef ui_entry;
-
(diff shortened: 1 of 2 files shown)


kolunmi:
awesome, merging this for now

PR #777, opened 2025-12-07
Update Spanish translations for software terms

DinimixisDEMZ:
"Libre" != "Gratis" in Spanish

- **Gratis**: Means 'zero cost'.
- **Libre**: Means 'freedom' or 'liberty'.

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -50,7 +50,7 @@ msgstr ""
 #: data/io.github.kolunmi.Bazaar.gschema.xml:11
 #: src/bz-preferences-dialog.blp:35
 msgid "Only Show Free Software"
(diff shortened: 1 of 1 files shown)


Cherryda:
You're absolutely right. Since it is a store I assumed it was the zero cost meaning, but Bazaar doesn't have priced apps yet as far as I'm aware, and Free is more used as in freedom in the FOSS ecosystem, so it's a huge oversight from my part.

DinimixisDEMZ:
>  it's a huge oversight from my part.

Nah, don't worry. These things happen all the time. The important thing is that you know now. 🧑‍🔧

kolunmi:
thank you so much! muchas gracias! <3

PR #781, opened 2025-12-07
Use WebP for screenshots originating from the Flathub CDN

AlexanderVanhee:
Both AppStream and the Flathub API seem to serve mostly the same image links, with the only difference being the file extension at the end. By manually changing the extension, we can fetch most screenshots in WebP format, saving space and bandwidth for both our users and Flathub.


kolunmi:
awesome, thanks a lot

PR #785, opened 2025-12-09
screenshot description fixed in metainfo

andrzejroskowicz:
Screenshots on flathub have mixed up descriptions, Nucleus app says "home page" and vice versa. This should fix it


kolunmi:
Thank you

PR #791, opened 2025-12-11
Add Leftover User Data page

AlexanderVanhee:
This PR introduces a new page where users can view a list of apps that still have user data, while the app itself is no longer installed. The size of the leftover data is also fetched and displayed on each tile to improve the overall UX.

Since we already have fetch this info, it’s straightforward to display it on the full app view page as well.

Screenshot From 2025-12-11 14-28-20
Screenshot From 2025-12-11 14-28-14


kolunmi:
Wow this is amazing!! Will take a look in an hour or so

PR #800, opened 2025-12-12
Make graphs have support for touch

AlexanderVanhee:
Both the normal and world map graphs should now support touch input like you would expect them to.

#800 🎉


kolunmi:
:fire:

PR #803, opened 2025-12-13

Lumaeris:
No shitpost this time, I think this is cool to use somewhere: https://notbyai.fyi/


kolunmi:
oooo, will discuss

PR #806, opened 2025-12-13
(Enhancement) Replace repo start count with Flathub bookmark count.

AlexanderVanhee:
This PR uses the brand new Flathub API route to show the bookmark count instead of the GitHub star count.

The advantages of this are:

- Possibility of bookmarking from within the app in the future when we implement the Flathub account system.
- Not having to make requests to GitHub from the user’s PC
- Not having to have a preference because of GitHub rate limits
- Works for all Flathub apps instead of just apps hosted on GitHub
- Is more representative of how liked it is on Linux specifically

image


kolunmi:
Amazing, thanks <3

PR #808, opened 2025-12-14
feat: devel version uses devel icon

renner0e:
seems like a qt skill issue to me, this one didn't get fixed
image

image
image
image


kolunmi:
Thanks!

PR #817, opened 2025-12-16

Saltane:
No guard is safe from the Dragonborn who has learned invisibility spell


kolunmi:
:mage_woman:

PR #819, opened 2025-12-16
Add Flathub Login

AlexanderVanhee:
This PR adds the Flathub login flow and basic auth state.

The login flow is built around WebKitGTKs WebView. A dependency included with the Flatpak runtime. It's primarily chosen to allow us to parse the login code from the redirect URL, which we then use to generate the authentication token.

I also added the Google icon for when that login method should ever get re enabled.

There is also the auth state, which is responsible for serialization and deserialization of its state, saving it via libsecret, and automatically logging out on token expiry.

Nothing is done with the credentials yet, but I plan to add bookmark management and, hopefully someday the ability to actually make purchases...

image

image

image

image

### DIFF
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -704,12 +704,12 @@ bz_application_flathub_login_action (GSimpleAction *action,
 
   g_assert (BZ_IS_APPLICATION (self));
 
-  window      = gtk_application_get_active_window (GTK_APPLICATION (self));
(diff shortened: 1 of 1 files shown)


kolunmi:
This is so awesome!! Thank you so much for working on this. I'm blown away

Remember to follow the style conventions set by the rest of the codebase. I think I'll make a markdown document detailing some of the more important things to do minimize memory issues.

PR #822, opened 2025-12-18
Translations: add Greek (el)

Xarishark:
I translated all available strings. Sorry if I made any mistakes with the way I am providing this pull request as its my first time contributing this way. 

Feel free to request any change I might need to do to finalize the files.

VScode had all languages as edited after I was done with Poedit so I pushed only the LINGUAS and el.po file just to be sure there are no edits to the other files(as they contained the word fuzzy even tho I never touched them with the poedit app).

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -4,6 +4,7 @@ bg
 bs
 cs
 de
+el
 es
 fa_IR
 fr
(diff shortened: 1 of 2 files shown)


kolunmi:
Awesome, thanks so much!! :greece:

PR #824, opened 2025-12-18
Add Disability Pride flag

fat0troll:
This commit adds the disability flag, as described here (https://en.wikipedia.org/wiki/Disability_flag) as the optional theme for the global progress bar.


kolunmi:
oh nice! will test

kolunmi:
hey, @fat0troll  I love this but the flag seems upside down from the reference image on wikipedia? Is this intentional?
image
image

shouldn't green be on the top? Apologies for my ignorance

kolunmi:
:thinking:

Lumaeris:
Just so you know, kolunmi haven't actually merged your PR, that commit belongs to entirely different PR. It seems that you had Pull app installed in your repo, and since there was some changes in upstream repo the app just force pushed all the changes from it to your repo, basically undoing everything you did here.

Either you need to disable it for this repo or create a new branch.

fat0troll:
ah, okay, lemme fix that

fat0troll:
> shouldn't green be on the top? Apologies for my ignorance

I think the flag is reading from left to right, and in the description of the flag red comes first, but dunno either, it's a diagonal after all. 

Would it be more logical to make green on the top?

PR #830, opened 2025-12-19
Add Favorites Page

AlexanderVanhee:
This PR adds the favorite page somewhat similar to https://flathub.org/en/my-flathub, it's based on the user data and install pages.

It should hopefully now not be too hard to add the favorite button to the full view.

Screenshot From 2025-12-19 12-57-24
Screenshot From 2025-12-19 12-56-04


kolunmi:
Absolutely incredible. The only issue I see is that the install/remove button does not become insensitive when you click it on the favorites page, meaning you could theoretically press it more than once while a transaction is already happening

kolunmi:
:partying_face:

PR #833, opened 2025-12-20
chore(po): add french for favorites

theMimolet:
Haiii,

I've added french translations for the new favorites page..
and..
that's it ! 

I also wanted to ask if there was a release schedule, like if it happens on specific days ? 
I feel kind of bad of "spamming" the translation commits when you add new features, so it would be convenient to know when an update is planned (not specifically a date, but like a day of the week or a frequence) so I can submit all of it the day before or something like this.

Note : i'm going to put more clarification in the title from now on because "update french translation" is too vague

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -10,8 +10,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: bazaar\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


kolunmi:
Don't worry, feel free to submit as many PRs as you'd like. There is no set release schedule, it's usually just when I feel like Bazaar is ready for one.

Thank you for the PR!

PR #834, opened 2025-12-20
Change the "Install Now" button string

yakushabb:
The dialog title says "Updates Are Available," but the suggested button says "Install Now," which is a bit confusing. This patch changes "Install Now" to "Update Now."


kolunmi:
thanks!!

PR #835, opened 2025-12-20
Make favorite button actionable

AlexanderVanhee:
The final part of the login experience for now is making the “favorites” button actually work for logged in users like you would expect.

If the user is not logged in, a popover is shown with an explanation. I don’t really care about pushing users to log in with an account, but simply disabling the button feels like terrible UX.

https://github.com/user-attachments/assets/7686db7a-fdaa-42fd-9f60-a0c029c7d886

Screenshot From 2025-12-20 21-49-30

### DIFF
--- a/src/bz-favorite-button.c
+++ b/src/bz-favorite-button.c
@@ -370,17 +370,13 @@ toggle_favorite_fiber (BzFavoriteButton *button)
     request = g_strdup_printf ("/favorites/%s/add", app_id);
 
   if (button->is_favorited)
-    {
-      dex_await (
-          bz_query_flathub_v2_json_authenticated_delete (request, token),
-          &local_error);
-    }
+    dex_await (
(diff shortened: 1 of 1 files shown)


Rexogamer:
apologies if this was already discussed, but one thing that I can see tripping people up is that - to me at least - the bookmark would suggest saving it for later rather than favouriting it? I feel like a star is more fitting but I might be missing something so feel free to correct me

kolunmi:
@Rexogamer This is a good consideration, but Flathub currently represents it as a bookmark. So for now I think this icon is fine

kolunmi:
@AlexanderVanhee this is shrimply amazing ( :shrimp: ), thanks so much as always!!

AlexanderVanhee:
I was thinking about submitting a PR to the Flathub monorepo to change their icon to a star, but they might not want that, since they don’t show favorite numbers publicly, so a bookmark icon makes more sense for them.

PR #848, opened 2025-12-25
Add "only show verified" option

AlexanderVanhee:
Some users prefer to only install verified apps. This filter option caters to that preference and was also easy to implement given our current filter mechanism.

I might also add mobile filtering, but I need to think of some way to only show this option on mobile to reduce preference clutter where possible.

image


kolunmi:
Looks amazing, thanks as always

PR #849, opened 2025-12-25
Make installed badge look better with Flathub style

AlexanderVanhee:
The green clashed to much with the purples of the Flathub page in dark mode.

image


kolunmi:
ooo yeah that looks better

PR #850, opened 2025-12-25
Show app page instead of search page when using search provider

AlexanderVanhee:
Added a new action to show an entry group on the full view with a specific app ID instead of just searching for that app ID.

I have changed it for the GNOME Shell search provider, but I do not know how to do it for the KRunner plugin. However, it should not break anything and it should be as simple as changing the used action name.

[Screencast From 2025-12-25 21-28-13.webm](https://github.com/user-attachments/assets/d743e8db-d6e2-4319-b7ab-9f5f771eb1c3)

### DIFF
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -629,6 +629,28 @@ bz_application_search_action (GSimpleAction *action,
 }
 
 static void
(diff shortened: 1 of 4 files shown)


kolunmi:
looks good

PR #852, opened 2025-12-26
Add an all apps in category page

### LINKED ISSUE #484: Category pages truncated at 96 apps - missing Flathub entries / no pagination
**Summary**
Category pages currently show a maximum of 96 app entries (24 rows × 4 columns) per category page. This limit does not match the full set of apps available on Flathub, preventing users from discovering all apps that should be accessible from the category page.
Is it a problem that can be solved ?

**Expected result**
Category pages should list all the apps present on Flathub that belong to that category (or provide a reliable pagination/search mechanism) so users can find every matching app. Use infinite scroll or arrows like on the gnome shell app grid.

**Actual result**
Only the first 96 apps are shown; additional apps present in Flathub are not visible or discoverable from the category page.

Thanks in advance

AlexanderVanhee:
This PR adds an "all apps" page which is used to show excess apps for category pages. It also adds categories to bz-entry and bz-entry-group.

Normal categories show the full listing of all apps with a specific category in their AppStream data, while pages like Trending or Popular only show excess entries previously shown on the normal page to improve performance on weaker hardware like mobile devices. The Adwaita category also shows its full ~500 apps.

An app listing page can either use GtkGrid or GtkListView for performant app listings that allow for basically infinite apps to be shown, or it can choose to contain other elements like the carousel (or even a simple label). That’s why this weird arrangement is needed.

The categories I had to pull into the entry and entry group might also be useful to create a search filter or something in the future, like seen on the Flathub site.

Closes #484

image


kolunmi:
wonderful, thanks

PR #858, opened 2025-12-28
Update size dialog

AlexanderVanhee:
This PR fetches the installed size and displays it in the installed list and the app size dialog. It also adds a "Details" section with additional information, similar to GNOME Software. This section will ultimately be moved to the currently unimplemented `SafetyContextDialog`.

image


kolunmi:
Cool!! thanks

PR #861, opened 2025-12-30
(UX) Add a fade to the releases list

### LINKED ISSUE #816: Release changelog field does not have a size limit
I believe this should be a field limited to like a single paragraph (something like 200 words max or something), then make it clickable into a "full screen" thing, similar to what we do with the image previews

Here is [drawpile's](https://flathub.org/en/apps/net.drawpile.drawpile) release description:

Image

Image

AlexanderVanhee:
Made it so the release description on the full view gets cut off like the app description, i decided to not add a "show more" button like GNOME Software as you can just as well click the releases history.

Fixes #816 

image

### DIFF
--- a/src/bz-releases-list.c
+++ b/src/bz-releases-list.c
@@ -22,6 +22,7 @@
 
 #include "bz-appstream-description-render.h"
 #include "bz-entry.h"
+#include "bz-fading-clamp.h"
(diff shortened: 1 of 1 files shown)


kolunmi:
yayy thanks :)

PR #867, opened 2025-12-31
Hide screenshot box for apps without any screenshots

AlexanderVanhee:
Also looked into hiding the placeholder app icon, but that didn't look too hot.

image


kolunmi:
good call

PR #875, opened 2026-01-01
i18n(el): fix translator credits  update translator name

Xarishark:
Corrects the translator credits entry in [el.po] to reflect the correct name (Xarishark-Zacharias Xenakis).


kolunmi:
thanks!!

PR #880, opened 2026-01-02
(Enhancement) Add app permission system

AlexanderVanhee:
This PR adds the app permission system and safety context dialog.

`bz-app-permissions` parses, creates and (de)serializes the permissions object.
`bz-safety-calculator` transforms the permission object, verified boolean and license into `safety-row`'s that are consumed by the safety dialog and full view.

I still have to change the icons seen on the full view to actually show what permissions are needed.

context-tiles-refresh-tbernard

Most of the code is just adapted from GNOME Software
Screenshot From 2026-01-02 15-25-30
Screenshot From 2026-01-02 15-25-25


kolunmi:
Nice, looks great!!! Thanks for putting in the time to do this :)

PR #887, opened 2026-01-02
fix: adding missing translations + updating Dockerfile

theMimolet:
Hi ! 

I added the login pages in the potfiles as well as the _All "Something"_ title -  I deleted "src/bz-all-apps-page.c" from it because it has no strings to translate in it. 

The only things missing are the pride color labels and the fact the webui for the login pages systematically goes to the english pages (even if the translated pages do exist). But both of these are technical problems rather than just adding them to the potfiles.

### DIFF
--- a/src/bz-preferences-dialog.c
+++ b/src/bz-preferences-dialog.c
@@ -162,7 +162,7 @@ create_flag_buttons (BzPreferencesDialog *self)
 
       button = GTK_TOGGLE_BUTTON (gtk_toggle_button_new ());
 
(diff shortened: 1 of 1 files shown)


theMimolet:
Welp at least I tried and the Dockerfile is good now ¯\(ツ)/¯

kolunmi:
Hi, thanks so much!

Could you remove your changes to bz-preferences-dialog.c ? They are not correct. I'll fix the issue right now in a separate commit

kolunmi:
thank you!!

PR #893, opened 2026-01-03
Refactor Context Dialogs

AlexanderVanhee:
Refactors the context dialogs to make them more consistent and better aligned with the code style guidelines.

Simplifies the code by adding shared widgets for the lozenges at the top and most action rows that all share a colored icon prefix, and removes dialog specific enums in favor of a more universal `BzImportance` enum.

Also fixes the age rating dialog context rows not being colored correctly.

image


kolunmi:
Looks amazing as always, thanks so much Alexander :)

PR #895, opened 2026-01-03
fix: make current Bazaar icons compatible with KDE Plasma

delphicmelody:
This will replace the current Bazaar icons with versions without clipping masks in order to ensure compatibility with KDE's QT SVG renderer as highlighted in the following issue.

https://bugs.kde.org/show_bug.cgi?id=448234


kolunmi:
Thank you, much appreciated!!! <3

PR #905, opened 2026-01-04
Trash User Data instead of deleting it

AlexanderVanhee:


kolunmi:
this will block the main thread, make sure to spawn a fiber on the io scheduler

AlexanderVanhee:
Switched to this https://gnome.pages.gitlab.gnome.org/libdex/libdex-1/integration.html

kolunmi:
That works as well, thanks

PR #916, opened 2026-01-05
Improve screenshot carousel to show more screenshots

AlexanderVanhee:
This PR makes an in tree copy of the carousel so we can improve it for our purposes. The current change removes the homogeneous spacing between the center of each widget, which previously caused screenshot carousels with different aspect ratio images to have massive gaps between screenshots.

To further show more screenshots, I changed the carousel so it starts from the second position when the user enters an app page on a wide window, avoiding a large empty space on the left side of the window. In that scenario, I also switched the first and second screenshots to make sure the best screenshot starts in the center.

I had to also fork the indicator dots as an Adw.Carousel cant be inherited from.

Screenshot From 2026-01-05 19-27-57

### DIFF
--- a/src/bz-carousel.c
+++ b/src/bz-carousel.c
@@ -57,7 +57,13 @@ struct _BzCarousel
 static void bz_carousel_buildable_init (GtkBuildableIface *iface);
 static void bz_carousel_swipeable_init (AdwSwipeableInterface *iface);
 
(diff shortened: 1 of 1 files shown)


kolunmi:
hmm, it seems the carousel doesn't follow focus?
[Screencast From 2026-01-05 12-30-05.webm](https://github.com/user-attachments/assets/4bd8d289-8cce-4520-ab27-b72b39d0d546)

AlexanderVanhee:
The Adw.Carousel also does not seem to follow focus.

kolunmi:
hmmm k

PR #921, opened 2026-01-06

theMimolet:
Hellooo,

Here's the updated french translations ! (Oh man so many strings :dizzy_face:)


kolunmi:
:fr: :heart_eyes:

PR #924, opened 2026-01-06
FlathubCategorySection: Make it more adaptive

Kekun:
This ellipsizes some labels to make the app better fit on mobile.


kolunmi:
Awesome!! Thanks so much for this

PR #925, opened 2026-01-06
ReleasesList: Wrap the fallback label

Kekun:
This wraps the fallback version information label to help the app fit on mobile.


kolunmi:
nice, thanks

PR #932, opened 2026-01-07
(Bug) Show licenses without a valid link in the license dialog

AlexanderVanhee:
Licenses that previously did not have a valid link like projects using a multi-license used the fallback text. Now it just displays the license text without formatting the license as a link.

This also fixes the text alignment issue.

image


AlexanderVanhee:
Should be fixed

kolunmi:
Awesome, thanks!!!

PR #933, opened 2026-01-07
(UI) Make Installed search rounded

AlexanderVanhee:
Also blocks global search on the installed page.

Screenshot From 2026-01-07 12-20-00


kolunmi:
Looks great, thanks!!

PR #935, opened 2026-01-07
Serialize spotlight and total entries for Flathub categories

AlexanderVanhee:
It caused the Trending, Popular, etc. sections on the Flathub page to briefly show titles before the API requests finished, resulting in an ugly layout shift.


kolunmi:
looks good, thanks

PR #940, opened 2026-01-08
Hide EOL by default

AlexanderVanhee:
EOL apps most likely have security issues and sometimes don't even have any metadata. So it's probably best to hide them by default.

### DIFF
--- a/data/io.github.kolunmi.Bazaar.gschema.xml
+++ b/data/io.github.kolunmi.Bazaar.gschema.xml
@@ -2,7 +2,7 @@
 <schemalist gettext-domain="bazaar">
   <schema id="io.github.kolunmi.Bazaar" path="/io/github/kolunmi/Bazaar/">
     <key name="hide-eol" type="b">
-      <default>false</default>
+      <default>true</default>
       <summary>Hide EOL Entries</summary>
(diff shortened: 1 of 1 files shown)


renner0e:
What about things like OBS that regularly have to use an EOL runtime?
I fear this could confuse users as they expect those popular apps to be present in bazaar.
Maybe the best compromise here would be to still show apps above a certain download count?

AlexanderVanhee:
Download statistics cant really be used as they need an individual API call to get the stats for each app.

OBS is also not marked as EOL in the app.

ghost:
There are times when one runtime is marked as EOL, but apps are not updated to the supported version.  

This behavior might be problematic.

EOL runtime does not always mean insecure. For example: games.

I am also working on this issue. 

https://discourse.flathub.org/t/eol-runtime-working-team/6627/12

ghost:
For example, at the moment:
- 190 apps use EOL versions of GNOME runtimes.
- 262 apps use runtime version 49, which will be EOL in a few months.  
- 512 apps use the latest version of the GNOME runtime.

ghost:
Sorry, I thought EOL means EOL runtimes. 

I have no idea Bazaar supports EOL apps!

PR #944, opened 2026-01-09
(UX) Add a no results page for installed apps

AlexanderVanhee:
Adds a status page for when the search query doesn't return any apps, like with the search page.

image


kolunmi:
:banana:

PR #946, opened 2026-01-09
Add "New Window" desktop action

AlexanderVanhee:
Since we do support multiple windows, it makes sense to enable this option for users who want to compare multiple apps or something.

image


kolunmi:
Thanks!!!

PR #949, opened 2026-01-10
Add separate warning for the X11 permission

AlexanderVanhee:
The full disk access warning didn’t make much sense for apps that only have the X11 permission. If an app has both permissions, the full disk access warning will still be shown as before.

Screenshot From 2026-01-10 14-16-21


kolunmi:
thanks!

PR #950, opened 2026-01-10
Update Dutch translations

AlexanderVanhee:
Also makes some string changes to improve the translation experience. Like removing translatable strings only visible to developers.


kolunmi:
:steamhappy:

PR #965, opened 2026-01-12
metainfo: add device support values

1peter10:
Thank you for creating this great app, I enjoy it very much - on desktop and mobile!

This adds the recommended values for apps that work on desktops and mobile
See https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#desktop-and-mobile-apps for reference. With this change, a future release of the app is going to be part of [Flathub's Mobile Collection](https://flathub.org/en/apps/collection/mobile/1).


kolunmi:
Thanks!! I appreciate it <3

PR #974, opened 2026-01-13
Add thumbnails for search

AlexanderVanhee:
Fetching lower resolution screenshots for search thumbnails reduces network usage and cache size and improves image loading speed.

I also made it so that screenshot variants from Flathub are preferred, as sometimes the first variant was a link pointing to some random site that doesn't even serve the screenshot anymore.


kolunmi:
looks good, thanks

PR #978, opened 2026-01-15
gettext: Add comments to Installed string

TitouanReal:
The string "Installed" is sometimes plural, sometimes singular. We need to add translator comments to make sure the string is correctly translated.

Here I marked one instance I found that I know is used for singular. Could I have a little help to know if the others are singular or plural? I don't know where to find them in the app.


kolunmi:
thanks

PR #981, opened 2026-01-16
Add "Install all" to favorites page

### LINKED ISSUE #869: [FEATURE REQUEST] Install all favorites button
Flathub has on it's page in "my Flathub" command to install all favorites (bookmarket) apps. Button in bazaar to install all favorites apps in one click would be very useful! Also maybe button to add all installed apps to favorites...

Image

Image

AlexanderVanhee:
This PR adds bulk install functionality. I tried to implement it in a way that can be easily reused for other bulk install use cases, like on the curated page.

Instead of creating a new dialog for bulk installs, I reworked the update dialog, since both are essentially just a list with some text and Cancel/Confirm buttons and should hopefully make a future redesign simpler to do.

I was also able to fix the issue where the favorites page would not show the empty page after removing the last favorite.

Closes #869

image

### DIFF
--- a/src/bz-transaction-dialog.c
+++ b/src/bz-transaction-dialog.c
@@ -570,9 +570,9 @@ bulk_install_dialog_fiber (BulkInstallDialogData *data)
   dialog = bz_transaction_list_dialog_new (
       G_LIST_MODEL (entries_store),
       heading,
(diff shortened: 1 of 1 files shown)


kolunmi:
Awesome, thanks

Issue #900, opened 2026-01-03
Flags not showing


linfindel:
### Describe the bug

The progress bar flags aren't showing either in the preferences or in the progress bar. Changing the selected flag doesn't help.

<img width="992" height="1017" alt="Image" src="https://github.com/user-attachments/assets/e1d98219-2cee-48de-87c8-4d516aee7ad2" />

<img width="622" height="448" alt="Image" src="https://github.com/user-attachments/assets/0602267d-3148-47e2-bcea-2de0c97cb9bb" />

<img width="622" height="448" alt="Image" src="https://github.com/user-attachments/assets/7745a3c5-cc33-49ec-beab-648a6c8a04a6" />

### OS

Project Bluefin, GNOME 49

### Version

0.7.2 (vcs=https://github.com/kolunmi/bazaar/commit/8fa8219299e7c2eb7660d1243c5ea56e0903205c)

### Configured Flatpak Remotes

Name              Options
Flathub           System

kolunmi:
strange, cannot reproduce. Will look into later

theMimolet:
Yup, I got the same issue on the same Bazaar version but on Aurora (KDE)

AshtakaOOf:
Running flatpak Bazaar from a terminal I get these errors:

```
WARNING: Glycin running without sandbox.

(bazaar:2): Gtk-WARNING **: 16:03:25.465: Theme parser error: style.css:1:55-57: While expanding --flag-gradient: Unit is missing.

(bazaar:2): Gtk-WARNING **: 16:03:25.465: Theme parser error: style.css:1:55-57: Unit is missing.

(bazaar:2): Gtk-WARNING **: 16:03:25.465: Theme parser error: style.css:1:54-56: While expanding --flag-gradient: Unit is missing.

(bazaar:2): Gtk-WARNING **: 16:03:25.465: Theme parser error: style.css:1:54-56: Unit is missing.

...
```

It repeats many times but I'm not gonna copy paste redundant logs.

linfindel:
They're randomly back this morning 🤔. I'll keep this open for now unless it's fixed for everyone else as well.

<img width="801" height="400" alt="Image" src="https://github.com/user-attachments/assets/fd26af89-814d-4c94-bf6d-5c3812118e4d" />

theMimolet:
Nope, still broken for me :<

NotoriousNico:
[...]

PR #983, opened 2026-01-17
Fix float formatting in pride flag css generation

marketagaldova:
See https://github.com/kolunmi/bazaar/issues/900#issuecomment-3763008053

`%f` uses the current locale to format the number, which may not be safe for code generation. For example, it may use a comma for the decimal separator, which is not valid in css. As far as i can tell, `%d` doesn't have this problem and rounding to whole percent values doesn't create a perceivable difference in this case.

This is perhaps not the cleanest way to solve this, but the other options as far as i can see are either explicitly using a safe locale, which wouldn't be thread safe i think, or some custom float formatter that always produces the correct output. Neither of these seem worth the effort to me. Please do correct me if im wrong tho, im a little out of my depth here.

### DIFF
--- a/src/progress-bar-designs/pride/pride.c
+++ b/src/progress-bar-designs/pride/pride.c
@@ -149,7 +149,7 @@ bz_get_pride_style_provider (void)
                           break;
                         }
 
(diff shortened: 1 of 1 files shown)


kolunmi:
Nice catch! Tysm!!

PR #990, opened 2026-01-19
Fix leftover user data bar being shown prematurely

AlexanderVanhee:
Fixes the papercut where the bar would be shown for a short time when choosing to remove all user data, allowing users to remove user data twice.

[Screencast From 2026-01-19 19-31-02.webm](https://github.com/user-attachments/assets/c5c7b5d5-aa9c-4428-ae0d-398d2e34d546)


kolunmi:
lgtm

PR #991, opened 2026-01-19
Have a rotation toggle for flags

AlexanderVanhee:
Now "horizontal" styles get generated automatically in `pride.c`

image


kolunmi:
AMAZING, thanks. Keep in mind, you only need to call `g_signal_handlers_disconnect_by_func` once on an object for a certain pairing of `func` and `user_data`

PR #993, opened 2026-01-19
Update Russian translation

fat0troll:
The term "недееспособный", used for Russian translation of Disability flag, is offensive towards the local disabled community members. Changed it to a more appropriative, widely used in Russian-speaking countries, term.


kolunmi:
thank you, good catch

Lumaeris:
Oh right, thanks for this! I wasn't aware that it was offensive in any way, good catch.

PR #995, opened 2026-01-20

Lumaeris:
the year has just begun and I already want it to end.

Also for some reason POEdit breaks the file to the point that GTranslator can't save it properly giving an error "format specifications in 'msgid' and 'msgstr' for argument 1 are not the same". Lokalize works for now

UPD: Apparently it because I've missed `%u` and `%d` somehow


kolunmi:
:( thx

PR #998, opened 2026-01-20
Remove Install and remove buttons for Bazaar

AlexanderVanhee:
Screenshot From 2026-01-20 20-54-27
Screenshot From 2026-01-20 20-54-33
Screenshot From 2026-01-20 20-54-40


kolunmi:
:+1:

PR #999, opened 2026-01-21
(UI) Improve stat dialogs

AlexanderVanhee:
Changes the tooltips in the two stats dialogs to use widgets, allowing for better layout and styling. Also updates the graph to no longer draw a raster, as it is now redundant with the tooltip and also kinda looked bad.

Screenshot From 2026-01-21 14-02-38
Screenshot From 2026-01-21 14-02-47


kolunmi:
ooo nice

PR #1013, opened 2026-01-24
(UX) Improve the transaction dialog

AlexanderVanhee:
This PR improves the transaction dialog by fetching remote metadata.

I also decided to make entries that cannot be acted upon insensitive instead of hiding them, as hiding them can be confusing.

Instead of fetching repository info for each entry, I decided to fetch the list once at startup and store it in the app state. This could be used to determine whether the (not yet implemented) scope selection preference should be shown or not based on if the user has multiple remotes.

Screenshot From 2026-01-24 19-39-21

Screenshot From 2026-01-24 19-39-17


kolunmi:
:steamhappy:

PR #1014, opened 2026-01-24
Add subcategories

AlexanderVanhee:
Adds subcategories using data from the Flathub API, they are fetched in a similar way to tags on app pages. Currently only the "Audio & Video" and "Gaming" categories have subcategories, matching the Flathub website. Other subcategories are apparently pretty crap.

image


kolunmi:
:justagirl:

PR #1027, opened 2026-01-25
docs: expand on curated section with the flatpak a bit more

renner0e:
preview: https://github.com/kolunmi/bazaar/blob/curated-docs/docs/overview.md


kolunmi:
yay! thanks

PR #1028, opened 2026-01-25
chore(po): updated French translations

theMimolet:
Hi !!

Saw that an update was about to be released,
So here's the express french refresh from yours truly :+1: 

Have an amazing night :D


kolunmi:
Thanks!! <3

PR #1029, opened 2026-01-26
Add read support for flatpak bundles

### LINKED ISSUE #984: Add cli option to preview metainfo
It would be nice if Bazaar had a cli option to allow users to pass in a metainfo xml file and preview what it would look like in Bazaar. 

Similar to the `--show-metainfo` option in gnome-software.

AlexanderVanhee:
Allowing users to open Flatpak bundles in read only mode should be a good first step toward future install support. It’s also useful in its current state for users who want to preview how their flatpaks metadata looks in Bazaar.

This also moves AppStream parsing out of the Flatpak entry.

Closes #984 

Screenshot From 2026-01-26 11-48-22

### DIFF
--- a/src/bz-releases-list.blp
+++ b/src/bz-releases-list.blp
@@ -1,8 +1,9 @@
 using Gtk 4.0;
 using Adw 1;
+using Gio 2.0;
 
 template $BzReleasesList: Adw.Bin {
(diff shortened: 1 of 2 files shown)


kolunmi:
:fire:

PR #1032, opened 2026-01-27
Updated pt_PT translation

azenyr:
Updates the pt_PT translation po file to try and keep it more up to date with current message strings.


kolunmi:
Thanks! :portugal:

PR #1036, opened 2026-01-28
Add support for single app remotes

AlexanderVanhee:
Adding support for this remote type will make the installed list more complete for users. It also allows users to remove or open these apps which is nice. This is also an important step toward ultimately supporting installation from bundles.

Entries from these remotes cannot be reinstalled, so they also do not appear in search.

image


kolunmi:
<3

PR #1039, opened 2026-01-28
Fix data graph not properly following accent color

AlexanderVanhee:


kolunmi:
thanks, good catch

changed `g_signal_connect` to `g_signal_connect_swapped` for consistency

PR #1040, opened 2026-01-28
Fix carousel background color applying to whole app

AlexanderVanhee:
Since adding CSS to individual widgets is deprecated, I have to use some kind of ID system to prevent these styles from overwriting each other when browsing between apps with multiple windows.


kolunmi:
:+1:

PR #1043, opened 2026-01-29
Add installed version support

AlexanderVanhee:
Should also be useful for displaying version changes in the update dialog in the future.

image


kolunmi:
cool , thanks

PR #1065, opened 2026-02-04
Increase paragraph spacing

### LINKED ISSUE #968: Appdata description and p tags
Appdata description tags and the related p tags appear to be interpreted slightly differently from GNOME Software. 

### GNOME Software

Image

### Bazaar

Image

AlexanderVanhee:
increased the paragraph spacing by 4 pixels, it’s still less than GNOME Software though.

@bertob Does this look good to you, and should we also increase the spacing between list items as well?

Closes #968 

Screenshot From 2026-02-04 20-59-31


bertob:
Could go a little bit higher (e.g. 12), but more or less looks good to me.

PR #1070, opened 2026-02-06
(UX) Library Improvements

AlexanderVanhee:
This PR tries to fix some rough edges with the library page:
- Adds a fallback image for update rows
- Merges runtime and add-on updates into a single row, as these often didn’t have proper metadata and therefore looked bad.
- Skips the .Locale, .Debug and .Sources refs for updates as these get updated with their main ref anyways.
- Stops trying to spawn the comet animations, which error out anyway
- Updates the transaction card so it no longer changes height when switching from and to the progress bar.

Screenshot From 2026-02-06 20-59-28

### DIFF
--- a/src/bz-transaction-tile.blp
+++ b/src/bz-transaction-tile.blp
@@ -34,14 +34,13 @@ template $BzTransactionTile: $BzListTile {
               "transaction-tile-title",
             ]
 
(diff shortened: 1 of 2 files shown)


kolunmi:
awesome. thanks!!

PR #1080, opened 2026-02-09
chore(po): update Indonesian translation

KiKaraage:
- 150+ new translations
- fix(age-rating): use "diperbolehkan" instead of "cocok"
- fix(mobile): use "Di Mana Saja" instead of "Dalam Genggaman"
- update strings for app categories, licenses, permissions & safety calculators
- update strings for newly improved library page
- update strings for preferences & (new) donation dialog

image


kolunmi:
yay thanks!! :indonesia:

PR #1081, opened 2026-02-09
Add exceptions for EOL filter

AlexanderVanhee:


rendering0:
Why add these 2 apps to the exceptions if users can simply disable the “Hide EOL apps” feature in the settings?

kolunmi:
We're not going to make exceptions for projects in favor of a configurable grace period after a project's runtime becomes EOL

kolunmi:
since @AlexanderVanhee  seems to disagree, I implemented this pr but with a config option:

https://github.com/kolunmi/bazaar/commit/be992815e9a687780a15cfdd78a37a9f1c2c14c5

kolunmi:
so adding

```yaml
override-eol-markings:
  - com.obsproject.Studio
  - net.lutris.Lutris
```

to the main config will do the same thing

PR #1084, opened 2026-02-09
chore(po): update & fix zh_CN translation 

jiesou:
The previous zh_CN translation has some issues and typos, for example:

- "Delete Data" mistranslated to "shan"
image

- "On the Go" mistranslated to "正在处理"
image

- The mixed use of "卸载" and "删除"

- Many expressions that were inconsistent with flathub.org

Changes I made:
- Fixed previous typos
- Unified terminology
- Added TRANSLATORS-ZH-CN.md
- More translations
- Many expressions are now aligned with the zh_CN translations on flathub.org

For example:

image

fixed and aligned with flathub.org
image

### DIFF
--- a/TRANSLATORS-ZH-CN.md
+++ b/TRANSLATORS-ZH-CN.md
@@ -82,7 +82,7 @@ msgmerge --update --verbose po/zh_CN.po po/bazaar.pot
 
 ## 测试你的翻译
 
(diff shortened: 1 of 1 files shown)


kolunmi:
awesome explanations! thanks so much

PR #1093, opened 2026-02-11
Hide title on app page until header is no longer visible

AlexanderVanhee:
It was a bit awkward seeing the app name twice so close to each other

https://github.com/user-attachments/assets/c51f9d5d-d9de-45c4-aaac-71b00f7ca345


kolunmi:
this is cool asf

Issue #726, opened 2025-11-27
Proxy Support


tronta:
Currently, if one is behind a proxy, e.g. a cooperate environment, Bazaar is not able to connect to flatpak.org.

Would be cool, if it respect the environmental variables...

tulilirockz:
How are you setting the environment variables? Could you take a look at https://flathub.org/en/apps/com.github.tchx84.Flatseal or [`flatpak-override`](https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-override)?

It seems that your environment is somehow misconfigured

tronta:
We use at work the proxy detox which sets the following envs:

NO_PROXY=127.0.0.1,localhost,::1
https_proxy=http://127.0.0.1:48157
no_proxy=127.0.0.1,localhost,::1
ftp_proxy=ftp://127.0.0.1:48157
http_proxy=http://127.0.0.1:48157
HTTPS_PROXY=http://127.0.0.1:48157
HTTP_PROXY=http://127.0.0.1:48157
FTP_PROXY=ftp://127.0.0.1:48157

So imho it would be great if those are picked up and used, or do I have a misunderstanding?

VirtualFreeEx:
Hi there, I just submitted a pull request with this feature. Thank you for reporting this, it pointed me in the direction of fixing the issue. ❤️

tronta:
Hi @VirtualFreeEx, thank you very much for your work.
I tried the latest version at work (7.10) and unfortunately the problem seems still to persist.

PR #1098, opened 2026-02-12
feat: Support proxy server resolution

VirtualFreeEx:
This used to be an issue in corporate *and* censored environments, such as Russia, which case I will go over now: `arewelibadwaitayet.com` is blocked over here, and the only way to use it is either proxying the request or modifying the packet in such a way that the DPI systems fail to detect it, which needs to be done system-wide and is generally hard(er) to achieve, especially compared to proxying network requests.

Additionally, an issue exists for a user requesting proxy support, which is needed for them in a coporate environment. Reference @ Issue 726.

Fixes and/or closes issue #726


kolunmi:
Hey thanks for this pr!! there seems to be a bunch of superfluous comments in this diff, could you remove those to be more consistent with the rest of the codebase?

VirtualFreeEx:
> Hey thanks for this pr!! there seems to be a bunch of superfluous comments in this diff, could you remove those to be more consistent with the rest of the codebase?

Done!

kolunmi:
hey @VirtualFreeEx , I decided to implement the better singleton-like approach you mentioned in your comment. Can you ensure everything is still working on your side?

VirtualFreeEx:
> hey @VirtualFreeEx , I decided to implement the better singleton-like approach you mentioned in your comment. Can you ensure everything is still working on your side?

~~Yep, I could. Let me know where I can access the patch/commit/diff to apply the changes locally and verify that everything still works.~~
nvm, missed the commit in commit log

VirtualFreeEx:
The feature still works, same as it did previously. Thanks!

PR #1100, opened 2026-02-13
Add Belarusian l10n

PlagaMedicum:
Hi!

Added full Belarusian translation. Made fully manually. Some strings are labeled as fuzzy, cause they could have more polishing in future (with terminology in particular).


kolunmi:
Amazing!! thank you so much :belarus:

PR #1107, opened 2026-02-13
Disable the update banner on first start

AlexanderVanhee:
It does not make sense for this banner to be shown to users who probably have not even used the app yet.


kolunmi:
:+1:

PR #1111, opened 2026-02-14
No installed entry selection

AlexanderVanhee:
Most of the uses were utilized to choose an entry from a group for removal or installation. However, since we already have `try_transact` with the nicer selection dialog UI, it makes more sense to just pass the entry group directly.

The only place where the removal is not yet possible is in the add-ons dialog, as it expects a single entry. And we cant just pass the UI entry as it would then be impossible to manage add-ons for the other entries.


kolunmi:
awesome, thanks

PR #1114, opened 2026-02-15
(UX) Improve error handling

AlexanderVanhee:
Instead of instantly showing an error dialog, this PR makes it so a toast with some human-readable text gets shown with a "Details" action that spawns the redesigned error dialog.

image
image


kolunmi:
Fantastic, thanks

PR #1118, opened 2026-02-15
docs: explain how to get useful crash reports

renner0e:
The flatpak install command is a bit unfortunate as this also pulls in
aarch64 and i386 Debug extensions, even if we don't need them at all,
but this is the most maintainable way to document this.

Or else we would have to update the documentation every time we update
the gnome runtime. As we only actually need
`io.github.kolunmi.Bazaar.Debug` and `org.gnome.Sdk.Debug`.


kolunmi:
:vlone:

renner0e:
image

PR #1123, opened 2026-02-16
Improve kw translation

linfindel:
On further thought, `klapkodh` is less ambiguous. Technically `kellgowser` can mean either a smartphone or a mobile dumb-phone (even if everyone only ever means the first). `klapkodh` is explicitly and exclusively a smartphone with apps and such :)


kolunmi:
Thanks!

PR #1133, opened 2026-02-17
(Enhancement) Update donation dialog

AlexanderVanhee:
I have made sure updating the release notes is as simple as pasting the notes from the appstream into the XML file.

Screenshot From 2026-02-17 15-10-23


kolunmi:
yaaay thanks :)

PR #1138, opened 2026-02-18
Use icon theme for installed refs

AlexanderVanhee:
This also fixes the issue were icons with only an SVG variant weren't being loaded.

### DIFF
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -663,9 +663,9 @@ bz_application_about_action (GSimpleAction *action,
                              GVariant      *parameter,
                              gpointer       user_data)
 {
-  BzApplication *self                    = user_data;
-  GtkWindow     *window                  = NULL;
-  AdwDialog     *dialog                  = NULL;
+  BzApplication *self   = user_data;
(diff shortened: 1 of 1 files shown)


src/bz-application.c
```diff
@@ -2595,6 +2595,23 @@ init_service_struct (BzApplication *self,
   bz_state_info_set_busy (self->state, TRUE);
   bz_state_info_set_donation_prompt_dismissed (self->state, TRUE);
 
+  {
+    GtkIconTheme    *user_theme        = NULL;
+    GtkIconTheme    *system_theme      = NULL;
```
kolunmi: please use g_autoptr() instead of manual g_object_unref

kolunmi:
Thanks, I just have one gripe which I commented above

AlexanderVanhee:
Where did you comments?

kolunmi:
@AlexanderVanhee 
image

AlexanderVanhee:
Ok, should be better now

kolunmi:
thanks

PR #1145, opened 2026-02-19
Strings Update

### LINKED ISSUE #922: Same translation to multiple fields even though sometimes it's not relevant
So it might not seem like an issue in english, but I just noticed that the potile is currently translating giving the same translation to multiple fields. 

By itself it's not necessarily an issue and they have the same msgid - but it has weird consequences for translations, here's a few example : 

> The Installed page is for example translated "Installées" in french -> app being female in french and plural in this context as the title refers to the multiple apps installed. However individual apps should be "Installé" or "Installée" with no plural - but currently there's no way to fix this.
Image

> In the taskbar, currently, the label for each task corresponds to the buttons name (Button : Install - Taskbar label : Install) however in french it might not be relevant, because we might want to put the verb "Installer" (To Install) for the button, while we might want to put "Installation" (Installing) for the label because it refers to the action in progress.    
Image

This is not a priority, but it would be nice if this was being taken into account.

theMimolet:
Hi ! 

There are some strings that seemed weird to me or not accessible enough.
So I've tried to fix that ^^"  

It's not finished, and I'm also going to fix #922 while I'm at it.
 
It shouldn't be included for 0.7.9 but 0.7.10 instead, so we don't need to delay it and redo the translations.
Feel free to give feedback on the changes !

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -11,8 +11,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: bazaar\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-02-22 13:26+0100\n"
-"PO-Revision-Date: 2026-02-22 13:31+0100\n"
+"POT-Creation-Date: 2026-02-22 19:06+0100\n"
+"PO-Revision-Date: 2026-02-22 19:07+0100\n"
 "Last-Translator: Mimolet <theMimolet@proton.me>\n"
 "Language-Team: French <>\n"
 "Language: fr\n"
@@ -809,8 +809,8 @@ msgid "Could not find app"
 msgstr "Impossible de trouver l'application"
 
 #: src/bz-application.c:3122
-msgid "Getting App Info…"
-msgstr "Réception des informations des applis…"
+msgid "Writing to cache…"
+msgstr "Écriture dans le cache…"
 
 #: src/bz-appstream-parser.c:303
 msgctxt "Project URL Type"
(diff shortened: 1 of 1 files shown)


AlexanderVanhee:
Could you use the special characters for the ellipsis please.
https://developer.gnome.org/hig/guidelines/typography.html#take-advantage-of-unicode

theMimolet:
Sir yes sir :saluting_face:

kolunmi:
Thanks for the pr!! Looks good, except for "No background tasks! -> "All your apps are up to date!": these do not mean the same thing. There can still be available updates when this label is active

theMimolet:
mmh I did not know that

theMimolet:
That's quite the commit so here's a list of stuff you might want to review : 
- Tried to replace strings that had "Synchronize" or similar with something I hope more accessible
- "Indexing data" sounded too technical, so I replaced it with "Getting App Info", but I'm not sure if it is what's really happening, I looked into the code a bit, found out it's linked to a "busy" state, but I still couldn't figure out what it did exactly - it's [here](https://github.com/kolunmi/bazaar/pull/1145/changes/cb3c0420fe74267018b229eb51dd3fae321830b8#diff-45ade5afced11aa0f2673e7a7bea5dbc0c539acd4afa34149aa6891a465038e3R3122)
- "Synchronize remote" sounded too technical too, so I replaced it by just "Refresh" (with CTRL + R, it also makes sense, along with "Reload")
- Linked to the previous change, I took the initiative of changing menu layouts to try to integrate "Refresh" better - here's how it looks like :
image
image
(here's how it looks in the "Manage user data" page, I forgot to change the layout there, if you want to compare :sweat_smile:)
image

If there's a change that's not okay, we can go back no problem
I'd be glad to hear your feedbacks, have a nice evening ^^

kolunmi:
Looks good!

kolunmi:
Thanks! Have a great day!

PR #1147, opened 2026-02-19
Proxy screenshots via Flathub

AlexanderVanhee:
This should hopefully reduce the cache size and the amount of data the user needs to download.

The non search thumbnail screenshots are less compressed so they still look reasonably good in the image viewer.


kolunmi:
nice, thanks!

PR #1150, opened 2026-02-20
chore(po): Update Polish translation.

Micro856:
This commit updates Polish translation, as well as fixes some mistakes I noticed when I were using the polish translation recently.


kolunmi:
Thanks! :poland:

PR #1151, opened 2026-02-21
Make holding shift while invoking a transaction skip confirmation dialogs

kolunmi:
This makes development easier and follows established paradigms, such as when you hold shift while pressing delete in a file manager, it removes the file permanently instead of placing it in the trash (sort of similar lol).


kolunmi:
:steamhappy:

PR #1156, opened 2026-02-21
Bring back transaction error button

AlexanderVanhee:
It was very idiotic that you had to hover over the text to see the error message, so I re added the error button, it now opens the new error dialog.

image


kolunmi:
I didn't even know you could see the error message by hovering anymore lol

PR #1160, opened 2026-02-22
Ask user to setup system Flathub when in Flatpak

AlexanderVanhee:
Newer versions of Flatpak support user remotes, but only for listing already installed apps, causing an edge case when users might only have the user Flathub. This results in the dialog not being shown, even though they cannot install apps from Flathub.


kolunmi:
Thanks! <3

PR #1163, opened 2026-02-22
Add ability to choose release notes version

AlexanderVanhee:


kolunmi:
Thanks for the pr. Not really necessary, and we may potentially forget to update this value in the future

PR #1167, opened 2026-02-23
Follow GNOME HIG's capitalization guidelines

ARAKHN1D:
Some labels and tooltips weren't following the GNOME HIG's suggestions on [capitalization](https://developer.gnome.org/hig/guidelines/writing-style.html#header-capitalization). If you disagree with any of these changes, I can adjust them.


kolunmi:
LGTM, thanks so much!

PR #1168, opened 2026-02-24
Update Czech translation

AsciiWolf:
Also fix one incorrect translation.

By the way, the following strings now appear to be duplicated (with inconsistent case):

```
#: src/bz-installed-tile.blp:77
msgid "Support this application"
msgstr "Podpořit tuto aplikaci"

#: src/bz-installed-tile.blp:101
msgid "Manage Add-ons"
msgstr "Spravovat doplňky"

#: src/bz-window.blp:402
msgid "_Login with Flathub"
msgstr "Přih_lásit se pomocí Flathubu"
```

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -11,7 +11,7 @@ msgstr ""
 "Project-Id-Version: bazaar\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2026-02-24 00:31+0100\n"
-"PO-Revision-Date: 2026-02-23 00:18+0100\n"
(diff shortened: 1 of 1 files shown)


AsciiWolf:
> By the way, the following strings now appear to be duplicated (with inconsistent case)

[...](https://github.com/kolunmi/bazaar/pull/1168#issue-3980750865)

(Not sure if this is intentional or not.)

/cc @ARAKHN1D

kolunmi:
thanks

ARAKHN1D:
That was not intentional, thanks for the heads up.

PR #1173, opened 2026-02-24
Fix some strings being duplicated

ARAKHN1D:
Oversight from #1167. I thought I caught them all there, but apparently not.


kolunmi:
No worries, thanks!!

PR #1179, opened 2026-02-25
chore(po): French refresh

theMimolet:
Just refreshing the some string that have been updated so what's already translated stays translated


kolunmi:
tysm! :fr:

PR #1187, opened 2026-02-27
Add mobile only preference toggle.

### LINKED ISSUE #1184: FR: Mobile-only toggle
I'm using Bazaar a lot on my OP6T with pmOS and one thing that I would love to see is a "Mobile apps" only toggle/mode. Of course this should be an optional setting.

It could be useful for some mobile distros to turn that setting on by default to offer a good experience by not recommending apps not mobile-compatible (they are a lot).

What do you think?

AlexanderVanhee:
This toggle is only visible when at mobile resolutions (or when its already enabled) to reduce clutter in the preferences menu. 
The Flathub page looks quite bad as many categories aren't filled, but this is luckily not an issue for mobile window widths where only 1 column is shown.

There is sadly no real better way to detect if you are on mobile.

image

Closes #1184


AsciiWolf:
Nice!

Huluti:
Thanks a lot @AlexanderVanhee, really good to see that happening!

AsciiWolf:
Are we still able to merge this for 0.7.11? :-)

AlexanderVanhee:
The current blocker is that the Flathub page looks bad with the toggle on due to lack of support.

image

Huluti:
Not so broken I think... Maybe "More ..." button could be hidden if there are no other apps to show?
And as you said the setting is meant to be shown only on mobile.

AsciiWolf:
So... Any chance this could be merged soon or is it still blocked by the bad Flathub page look? :)

AsciiWolf:
@kolunmi Sorry for the ping, but could I ask what is the state of this feature? Is it still blocked or is there a chance it could be merged soon?

I ask because it is probably one of the most important features to me since I plan to use Bazaar on mobile devices.

Thanks! :)

kolunmi:
@AsciiWolf  I should take another look at this, maybe it's not so bad anymore. Plus it would be useful as you said, even if we would want it off by default

radmorecameron:
"There is sadly no real better way to detect if you are on mobile."

I asked on Fediverse about this btw since I wanted some guidance on another project where functionality changed based on the device type 😄 

Using `hostnamectl chassis` was suggested https://social.librem.one/@dos/116770030426595448

It seems like you'd want to check if the chassis is "tablet" or "handset" in this case.

The only problem with min size that I can think of is if someone plugs their phone into a monitor then they might not see this toggle.

PR #1189, opened 2026-02-28
Update io.github.kolunmi.Bazaar.desktop.in

Dymkom:
Added Ukrainian localized keywords to allow finding the application when searching using Latin transliteration, improving and expanding overall search discoverability,


AlexanderVanhee:
This gets automatically added already from the .po file when building the app.

Dymkom:
@AlexanderVanhee, thank you for the explanation.

PR #1190, opened 2026-02-28
Refactor installation logic

AlexanderVanhee:
This PR moves the install button area into its own widget to avoid code duplication, it also allowed me to more easily add an update button there and to improve transitions.

I also removed a ton glue logic by using actions for install, bulk install, remove, show entry and addons. But not yet for update as this one uses an entry which may not have its result realized yet.

image


kolunmi:
absolutely beautiful. thanks

PR #1197, opened 2026-03-01
add malcontent support (parental controls)

AlexanderVanhee:
Add the malcontent service, which maintains a list of blocked refs and a maximum age rating value. This is used to restrict the launch of certain applications and to filter which apps are displayed in the UI respectively.

There is also a "Restrict Application Installation" toggle. However, this is already handled by libflatpak, which prompts the user to enter an administrator password when installing or removing applications if the toggle is on.

You will need an extra user account to test these changes out.

image


kolunmi:
this is great!! thank you for following the style rules <3

PR #1202, opened 2026-03-02
Fix age rating notification

AlexanderVanhee:
It was using the wrong variable binding (which happened to change at the same time), and I also forgot to add it to search.


kolunmi:
:justagirl:

PR #1207, opened 2026-03-03
chore(po): French refresh

theMimolet:
Same reason as last time
Refreshed the strings that were touched to make sure the translation is kept


kolunmi:
thank you!! :fr:

PR #1208, opened 2026-03-03
Full Bundle Support

kolunmi:
Right now seems to be able to load and open bundles, but installation simply asks for privaleges, then the transaction itself finishes instantly with no apparent effect


kolunmi:
testing with: [adw-demo.zip](https://github.com/user-attachments/files/25725310/adw-demo.zip)

kolunmi:
Here is a video of the issue: https://imgur.com/a/KVHKTGR

AlexanderVanhee:
Warehouse gives this error: 

error: The application org.gnome.Adwaita1.Demo/x86_64/master requires the runtime org.gnome.Platform/x86_64/master which was not found

Other bundles that depend on a non master runtime seem to work for me, like the one from the Hytale site.

kolunmi:
I wonder why bazaar doesn't show the error

we also need to clean up a few parts of UI for this before merging, like how you seemingly can't launch hytale from bazaar's UI afterwards

kolunmi:
superseded by https://github.com/bazaar-org/bazaar/pull/1424

PR #1227, opened 2026-03-07
Add layout manager to app tile for preferred width

### LINKED ISSUE #963: (UX) App tile flow boxes show different column counts depending on their content
I would like it to behave like GNOME Software, where flex boxes change the number of columns based solely on the window width, avoiding situations where different sections have different column counts at certain widths.

Image

Image

AlexanderVanhee:
This fixes the issue where the number of columns was inconsistent between different flow boxes at the same window width.

It also allows us to remove the character limits from the title and description.

This is almost directly ported over from GNOME Software

Closes #963 

image

### DIFF
--- a/src/bz-app-tile.c
+++ b/src/bz-app-tile.c
@@ -308,7 +308,8 @@ bz_app_tile_set_preferred_width (BzAppTile *self,
 
   self->preferred_width = preferred_width;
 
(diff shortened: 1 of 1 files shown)


kolunmi:
:steamhappy:

PR #1238, opened 2026-03-08
Adds support for links in description renderer

AlexanderVanhee:
image
image


kolunmi:
Wow, this is awesome!!! I will however ask that you convert this pr to use the various `g_utf8_` functions so that the code doesn't get confused with other languages that might not use ascii :bird:

AlexanderVanhee:
I think I removed all the ASCII dependencies

PR #1240, opened 2026-03-08
Make focus start on install or open button when opening full view

AlexanderVanhee:
This should fix the workflow where you just had to press enter to install an app when opening from the search provider.

The padding changes are there to fix a focus outline clipping issue caused by the stack.


kolunmi:
Thanks

PR #1243, opened 2026-03-09
desktop: List compatible form factors

fossdd:
List the form factors the app was designed for in the desktop file. This is so desktop environments like Phosh and Plasma Mobile can show or hide the app in the correct contexts.


kolunmi:
Awesome, thanks! Have a great day!

PR #1246, opened 2026-03-09
Force icon theme to "Adwaita" at start

AlexanderVanhee:
This PR adds a setting to force the icon theme to "Adwaita", which is enabled by default.

I believe this is the best approach due to the following issues that badly impact the user experience when making use of themes like Breeze (KDE) or Yaru (Ubuntu):

Some icons randomly fail to load. 
image

 Some icons do not convey the correct intent.
  For example, the © icon is supposed to be a set of scales here, indicating a license in general, not specifically copyright. 
image

Some icons appear randomly with their high-color variants, which looks terrible.
image

Only about half of the icons are actually themed, resulting in a mix of icon styles, primarily visible due to the difference in line thickness.

It can be disabled with the following command:
```
flatpak run --command=gsettings io.github.kolunmi.Bazaar set io.github.kolunmi.Bazaar force-adwaita-icons false
```

### DIFF
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -2802,12 +2802,15 @@ init_service_struct (BzApplication *self,
   self->settings = g_settings_new (app_id);
 
   if (g_settings_get_boolean (self->settings, "force-adwaita-icons"))
-  {
-    GtkSettings *gtk_settings = NULL;
-    gtk_settings = gtk_settings_get_default ();
+    {
+      GtkSettings *gtk_settings = NULL;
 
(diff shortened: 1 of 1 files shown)


kolunmi:
erm akshually this pr is infringing on my rights :nerd_face: 

way to go red hat...

AlexanderVanhee:
[Please do not theme this app](https://stopthemingmy.app)

dominichayesferen:
Saw this because it was being spread around - probably best to lock it 😅 
_If only GTK forced in an Adwaita fallback at least like KDE forces Breeze as an icon fallback..._

PR #1258, opened 2026-03-10
Add lingering cache to async textures

AlexanderVanhee:
Adds a hashmap containing references to loaded GdkTexture objects, keyed by source URIs. Used for deduplication of loads, but primarily to prevent textures from blinking when an BzAsyncTexture for the same URI is destroyed and recreated like half a second later.

### DIFF
--- a/src/bz-async-texture.c
+++ b/src/bz-async-texture.c
@@ -38,130 +38,35 @@
 #include "bz-io.h"
 #include "bz-util.h"
 
-typedef struct
-{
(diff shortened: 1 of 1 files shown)


AlexanderVanhee:
@kolunmi what should happen with this?

kolunmi:
> @kolunmi what should happen with this?

I like it just have to take a closer look

kolunmi:
please follow the style rules, they are there for a reason

PR #1264, opened 2026-03-12
Update kw translation

linfindel:


linfindel:
Hmm, why is the diff empty? 🤔

PR #1271, opened 2026-03-16
Add permission editing

AlexanderVanhee:
Adds Flatpak permission editing.

Features:
- Static (override keyfile) and dynamic (portal) permission editing
- Reset button for changes
- Human-readable descriptions for each permission
- Split between normal and developer-oriented permissions via a revealer
- Warning banner
- Entry points from the permission dialog & installed app tiles.

Notes:
- I decided to make the reading/writing of the keyfile synchronous, as the file should be tiny with very little blocking, and making it all async would bring much complexity.
- Global overrides seem like a bad idea to me as people could easily mess up almost all their apps.
- I hope to use animated Gtk.Svg from GNOME 50 runtime to improve the portal tiles.

Features I still want to add but did not make it in:

- Watching of the override keyfile
- Some kind of visual indicator on the installed tile making it obvious if it has edited permissions.

image

### DIFF
--- a/src/bz-permission-toggle-row.blp
+++ b/src/bz-permission-toggle-row.blp
@@ -5,32 +5,30 @@ template $BzPermissionToggleRow: Adw.ActionRow {
   activatable-widget: toggle;
 
   [suffix]
-  Button reset_button {
-    icon-name: "edit-undo-symbolic";
-    valign: center;
-    sensitive: false;
-    tooltip-text: _("Reset to default");
-    clicked => $reset_button_clicked_cb();
+  Box {
+    Button reset_button {
+      icon-name: "edit-undo-symbolic";
+      valign: center;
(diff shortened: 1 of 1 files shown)


AsciiWolf:
Nice!

P. S. I would probably omit the specific system names like PulseAudio or CUPS to make this not too confusing to less experienced users. These names are not needed anyway since CUPS is basically the only print server on Linux nowadays and PulseAudio or PipeWire are the only sounds servers.

Modifying the warning banner to mention that changing the default app permissions (and bypassing the sandbox) can put the system and user data in risk etc. could also be a good idea in my opinion.

kolunmi:
@AsciiWolf we'll refine the UI strings in smaller iterative prs

kolunmi:
@AlexanderVanhee  amazing! thanks!!

PR #1279, opened 2026-03-18
Make BzGroupTileCssWatcher apply responsive styling

kolunmi:
Generate css for `:hover` and `:active` states so it doesn't feel so static


kolunmi:
https://github.com/user-attachments/assets/8d9faee3-8fb4-43b9-b4d6-c453a2aa3663

PR #1281, opened 2026-03-18

theMimolet:
Hi ! 

Here's my French translation for the new permission page ! 
It's fully translated but it's still a draft because I need to ask my peers for their opinion for some strings I still have doubts on :face_with_diagonal_mouth:


theMimolet:
I also noticed it's using the same string for the "remove" label
One being used for app management (uninstall) and the other being used in bus policies (and more)
It would be nice to be able to differentiate between those two - especially when the translations depend on the context

theMimolet:
Sad about the permission part being scrapped.. 
Hope it might at least give ideas to Flatseal or Gnome because the design (especially the one of the top buttons) was really good.. 

Anyway, here's the french translation of the latest changes, finally ready :D

kolunmi:
Thanks! :fr:

PR #1283, opened 2026-03-19
Use the install_last_month JSON field from Flathub API

AlexanderVanhee:
No reason to calculate something ourselves that often does not even match with Flathub.


kolunmi:
thx

PR #1288, opened 2026-03-19
Add undo button to permission reset toast

AlexanderVanhee:


kolunmi:
Thanks! Make sure you are not performing IO in a way that blocks the main thread, like with g_mkdir_with_parents in a button callback. Pls spawn a fiber

AlexanderVanhee:
This whole file performs IO operations on the main thread, based on the assumption that the keyfile is small enough that the added complexity of making it asynchronous is not worth it.

kolunmi:
if you used libdex fibers on the main scheduler then it wouldn't be complex :vlone:

kolunmi:
this file needs modifications for that then tbh. If I keep on allowing this sort of thing too much it will eventually get out of hand as the style has. One of bazaar's main goals is to never do I/O on the gui thread unless it is absolutely necessary, and even then maybe like one call. 

I'll merge this for now

PR #1289, opened 2026-03-19
feat(ci): automate flatpak dependency updates

renner0e:
Currently dependency updates are completely manual effort on this repo,
automatic PRs are only made to the stable version on flathub, which
means that dependency updates are essentially not tested as they
usually go straight into the flathub manifest.

I made this so it won't be super annoying and only run on important
updates, just like in flathub. When a bazaar release is made changes
can be more easily carried over to the flathub manifest because of [1].

We can also turn off the data-checker on flathub so we only have one
place to get spammed with bot PRs and direct contributions to the
manifest here so they can be tested.

You will get a PR like this [2].

[1]: https://github.com/flatpak/flatpak-builder-tools/blob/master/flatpak-json2yaml.py

[2]: https://github.com/renner0e/bazaar/pull/5

### DIFF
--- a/build-aux/flatpak/io.github.kolunmi.Bazaar.json
+++ b/build-aux/flatpak/io.github.kolunmi.Bazaar.json
@@ -53,8 +53,7 @@
           "commit": "aa5298cc1677cf855bd61f9671607362bd203f4f",
           "tag": "v0.20.0",
           "x-checker-data": {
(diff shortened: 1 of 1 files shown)


kolunmi:
:vlone:

renner0e:
> :vlone:

image

PR #1294, opened 2026-03-20
Remove Permission editing

AlexanderVanhee:
:(


kolunmi:
:(

AsciiWolf:
Wait... Why?

Also, [this](https://github.com/kolunmi/bazaar/blob/main/data/io.github.kolunmi.Bazaar.metainfo.xml.in#L16) and [this](https://github.com/kolunmi/bazaar/blob/main/data/io.github.kolunmi.Bazaar.metainfo.xml.in#L74) should probably be removed as well.

PR #1300, opened 2026-03-21
Fix system Flathub not being added if user Flathub was found in sandbox

AlexanderVanhee:
The UI was changed to handle having a user Flathub and no system, but I forgot to update the `ensure_flathub_fiber`. So it just skipped adding the remote because it already saw the user one.

Also makes it so the Flathub page shows the "Flathub missing" page again when its supposed to.


kolunmi:
thx

PR #1304, opened 2026-03-22

AsciiWolf:
This time I decided to keep the old, commented out strings. This makes it possible to easily re-add them if the permission editing functionality gets restored in the future.


kolunmi:
Thanks!! :czech_republic:

PR #1307, opened 2026-03-22
Make runtime titles not have double versions

AlexanderVanhee:
For some reason, some GNOME runtime titles included a version number like "49" at the end, while the actual version was something different like 50. This caused the old algorithm to combine both values and return "49 50". To fix this, we now remove the version number from the title entirely and append the correct version value.


kolunmi:
It might be inefficient to compile that regex every time we process a new entry, perhaps we can wrap this in a `g_once_init_pointer`?

kolunmi:
thx

PR #1313, opened 2026-03-22
Add "Since" label to world map for context

AlexanderVanhee:
This mirrors what Flathub is doing.

image


kolunmi:
nice attention to detail lol

PR #1315, opened 2026-03-22
Improve world map loading mechanisms

AlexanderVanhee:
Makes it so the locale data for the world map gets automatically generated at build time and changes the output to GVariant for improved loading speed over JSON.

This seems to mostly stop the lag spike when opening the graph dialog.


kolunmi:
awesome, thanks!!!

PR #1319, opened 2026-03-22
(UX) Show open button on search tile after install

AlexanderVanhee:
Users are probably more interested in running the app they just installed than removing it, so it now shows an "Open" button until the search is changed or something.

Screenshot From 2026-03-22 21-36-59


kolunmi:
image

PR #1328, opened 2026-03-28

Micro856:
Updates Polish translation

P.S. Please merge it in a way that will allow me to discard commits on my fork as I can't, and it is not looking pleasant with these 20 commits.


kolunmi:
thanks! try using a new branch each time you make a pr

PR #1331, opened 2026-03-29
chore(id): Update Indonesian translation

KiKaraage:
Feat some good Eid cuisine from quite the last week - this one is opor ayam ketupat! (Chicken coconut stew with steamed rice cakes and chilli sauce)

image


kolunmi:
ooo yummy!! Thanks so much!!! :indonesia:

PR #1355, opened 2026-04-07
Make other apps by developer not use Flathub API

AlexanderVanhee:
This makes the time until result shown way shorter, saves an API call to Flathub and makes it work for apps across all remotes.

@kolunmi does the filtering need to be made async via dex or would the overhead be larger than the ~5000 string compares it currently has to do on the main thread?


kolunmi:
I'll take a look soon, thanks for this pr!!

PR #1358, opened 2026-04-08
Remake the global progress bar with wdgt

kolunmi:
WIP, see https://hachyderm.io/@kolunmi/116367952214637707


kolunmi:
https://github.com/user-attachments/assets/84bbf910-8c03-42c0-9308-a48ff1115607

kolunmi:
I'll just merge this now, if we don't like the look we can adjust the params later

AlexanderVanhee:
Sorry for the late review, I was still asleep.

The corner radius of the inner rectangle does not seem to match the outer one like it did with the old progress bar.

The more important issue is whether we want to use this new "wdgt" system. Does it bring enough value to justify its maintenance over just using Snapshot or the new Gtk.Svg? Could relying on this custom system alienate new contributors or jeopardize our future path towards possible inclusion into GNOME Core?

kolunmi:
@AlexanderVanhee  I think as long as we don't rely on it for critical stuff for now it will be fine. But overall I do feel this is something I've wanted to invest time into experimenting with and I'm pleased with how well it works to reduce boilerplate

kolunmi:
If you want to/have time submit a pr to fix the corner radius it should be pretty easy to figure out, otherwise I'll fix it in a bit. Good call

kolunmi:
1d3cad3

PR #1364, opened 2026-04-08

theMimolet:
Bonjour, French Crevette here :shrimp: 

Small french update (there was one new string) 

Hope you have a nice day ! :D


kolunmi:
Thanks!!!
copilot should be off now

PR #1393, opened 2026-04-11
Add support for previewing metainfo files

AlexanderVanhee:
Adds a command-line argument that opens two file selection dialogs, allowing the user to choose a metainfo file and optionally an icon file. It then displays a mock app page, including a banner with a button that leads to a basic page showcasing widgets elements that have their branding.

This should make it easier for developers to design their metainfo file based on how it will appear in Bazaar.

The file dialog approach was chosen because I couldn’t figure out a way to pass two different files into the sandbox via launch commands. It’s also a bit more user friendly, I guess.

`flatpak run io.github.kolunmi.Bazaar --preview-metainfo`

image
image

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -88,6 +88,7 @@ src/bz-license-dialog.h
 src/bz-login-page.blp
 src/bz-login-page.c
 src/bz-login-page.h
+src/bz-metainfo-preview.c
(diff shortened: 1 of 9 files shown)


kolunmi:
:vlone: looks good lemme test locally

PR #1406, opened 2026-04-12

theMimolet:
Hello !

Here we go again,
As a draft, once again, because there will be lots to translate >:)

Also thanks for clarifying the vocab on the graph, it was driving me nuts when I first tried to translate it ^^"


theMimolet:
I think I'll leave this PR there
If there are more strings that need translation, I'll open a new one

Maybe a note for UX : While improving older strings, I found out that in 

> "You have a network connection but are viewing a cached version of Flathub"

putting there "an offline version" instead of  "a cached version" might be better for a broad audience

Anyway, [I go to sleeeep](https://www.youtube.com/watch?v=6NlzcEtnfn4)
Byyyyee

kolunmi:
Thanks again for your help!! :fr: :sleeping:

PR #1409, opened 2026-04-13
Put progress bar behind the install button

kolunmi:
https://github.com/user-attachments/assets/f1e8ec5b-b0e5-4767-bd37-1ea0183ebe41

@bertob


AlexanderVanhee:
The button seems to lack a hover and active state, I would put the "Install" and "Cancel" text in a stack so the width of the button doesn't jump when the state changes. The inner bar is way to wide on mobile when in the pending state

kolunmi:
@AlexanderVanhee fixed the jumping problem. I've discovered more places wdgt needs improvements along the way :hamster:

kolunmi:
ok worky

bertob:
A few points of feedback from Matrix, so they don't get lost:
- Decrease the around padding of the inner pill in the initial loading state
- There's a 2-3s state at the end where the progress bar is full and static, which looks like a different style for the install button. Maybe we should stall a little bit before the end, and only transition to complete when we're ready to move on.
- When uninstalling an app there's a 2s delay before the buttons are replaced by the Cancel button, it'd be good to have immedate feedback instead

bertob:
A few more things I came across:

- The new progress button is not used on the Addons dialog
- By cancelling the install right as it's about to be complete you can end up in a limbo state where the app is (presumably) installed, but there's still the Install button rather than Open/Uninstall. In that state clicking Install does nothing, it just glitches for a second and then goes back to Install

[Screencast From 2026-04-17 13-13-49.webm](https://github.com/user-attachments/assets/eaa512a6-a305-46e7-9f09-c03ea87d2813)

bertob:
One more small nitpick: In the initial state where the inner pill is just going back and forth, I'd make that inner pill a bit less wide. Other than that and dropping the global progress bar this looks good to go from my side :)

notaf0e:
Another piece of feedback related to this. If the app is waiting to be installed in queue I believe the button should instead display "in queue" with either no progress bar, or instead bouncing from one side to the other like in the video at the start. Keep up the beautiful work! <3

kolunmi:
@bertob how's this look for the inner pill width?

https://github.com/user-attachments/assets/9d259ca7-c308-44cd-9f7e-e4f635bc6958

kolunmi:
I feel that perhaps there should be a fourth state for when the download gets cancelled

currently it waits for the backend to finish cancelling before showing anything, which can often be delayed if a network operation is ongoing

kolunmi:
@bertob  how do you feel about this for a "cancelling" state?


https://github.com/user-attachments/assets/7b232665-357e-45f5-9c4c-27d6b93d2b25


@notaf0e  I believe your concerns should be solved with 61879d4795f6742d1c045b6d54f36f979012bc64 as well

kolunmi:
Since we've achieved the main goal of this pr, I'll merge and we can make further adjustments later if necessary

AsciiWolf:
It seems this PR has broken the translation of the "Install" button: https://github.com/bazaar-org/bazaar/issues/1461

PR #1415, opened 2026-04-14
Fix updates for user installs in sandbox

AlexanderVanhee:
Flatpak would set the `.desktop` file to point to the Flatpak executable found in the Bazaar sandbox, so breaking the `.desktop` file. Overwriting the `FLATPAK_BINARY` environment variable with the true Flatpak executable path fixes this.

I made it so user update fetching gets disabled if it can’t find the correct path.

### DIFF
--- a/src/bz-flatpak-instance.c
+++ b/src/bz-flatpak-instance.c
@@ -757,6 +757,22 @@ init_fiber (InitData *data)
       g_clear_pointer (&local_error, g_error_free);
     }
 
+#ifdef SANDBOXED_LIBFLATPAK
+  if (g_getenv ("FLATPAK_BINARY") == NULL)
+    {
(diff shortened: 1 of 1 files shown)


renner0e:
I can confirm this fixes the issue under these conditions

```
flatpak install --user org.gnome.Calculator
flatpak update --user org.gnome.Calculator --commit=5e09c43a2b65723fd726685d08ec45f72e89669756ea9c07b440f385a027acc6 -y
```

refresh and click update in bazaar

image

kolunmi:
looks good! However `g_strstrip` returns dynamically allocated memory, you need to free

AlexanderVanhee:
https://docs.gtk.org/glib/func.strstrip.html Does it?

kolunmi:
o nvm :vlone:

PR #1418, opened 2026-04-15
Make exported desktop search provider app icon sizes adjustable

kolunmi:
Added a new environment variable `BAZAAR_DESKTOP_SEARCH_PROVIDER_ICON_SIZE` which may be read as an unsigned integer from 1 to 256 to specify the square pixel size of icons sent to the desktop search provider. This should be configured to be a higher value than its default value of 24x24 when on KDE, as krunner paints result icons as much larger than gnome-shell does. This way, the images will not appear blurry as they have historically on KDE.

@renner0e :eyes:

### DIFF
--- a/src/bz-entry.c
+++ b/src/bz-entry.c
@@ -2343,6 +2343,7 @@ GIcon *
 bz_load_mini_icon_sync (const char *unique_id_checksum,
                         const char *path)
 {
+  guint            icon_size             = 0;
   g_autofree char *main_cache            = NULL;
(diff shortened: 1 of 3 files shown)


renner0e:
Worky! Thank you so much, this is really nice.
image

image image

Results with:
`BAZAAR_DESKTOP_SEARCH_PROVIDER_ICON_SIZE=128 flatpak run io.github.kolunmi.Bazaar`

kolunmi:
image

PR #1421, opened 2026-04-16
feat: add minecraft search tweaks

renner0e:
This is a common abbreviation for minecraft, anything to do with it is already pretty popular on flathub anyway and this sorta thing has already been done for youtube.

### after:
image

### before:
image


kolunmi:
ooo thanks

I think your regex is matching both `m` and `mc` since there is a `?` after the c

image

kolunmi:
image

PR #1422, opened 2026-04-16
fix(just): build before bundle

renner0e:
build-bundle is gonna fail when you don't build beforehand, we can use disable-updates here which makes iterative building so much faster because we don't update git sources anymore each time.


kolunmi:
thx

PR #1424, opened 2026-04-17
add bundle install support

kolunmi:
image


AlexanderVanhee:
Current issues: 
- Only opens the page on the second attempt of opening a bundle
- Both bundle and remote entry gets added to the cache
- Removing just after installing does not update the "installed" property of the entry.

Would it be easier for the backend to just have a separate dialog to install the bundle? It's something that I want to do in the future anyway...

AlexanderVanhee:
It seems to work better now, but the bug where the remove button does not disappear when removing just after install still persists.

AsciiWolf:
Don't forget to add the _BzBundleInstallDialog_ to POTFILES.in. ;-)

kolunmi:
https://github.com/user-attachments/assets/c596bd82-0481-4f5a-96f4-6c2139e16f0d

now we just need to make the UI pretty

kolunmi:
Gonna make the UI mockup implementation a separate pr. For now I want the technical improvements in main

PR #1438, opened 2026-04-23
Refresh entries with a subprocess

kolunmi:
Introduces `bazaar-refresh-worker` which pulls remote flatpak repo content outside of the main bazaar process. Once this subprocess exits, the many many allocations made by the complicated refresh routines, especially inside libxmlb, can be reaped instantly in a way that is more effective than using `malloc_trim()`. This significantly reduces memory usage over time.


kolunmi:
the main bazaar process now sits at around 225-275 mb according to btop, even after many refreshes

AlexanderVanhee:
LGTM

PR #1450, opened 2026-04-28
Fix custom widgets having no outline in HC mode

zoeyAhmed:
See commits for each fixed component

|Before|After|
|--------|-------|
| image | image|
| image | image|
| image |image|
| image |image|


kolunmi:
oooo thanks!

PR #1456, opened 2026-05-04
Add GNOME Extension support

AlexanderVanhee:


kolunmi:
hmm in search results it is a bit hard to tell which is an application and which is a gs extension

src/bz-gnome-extension-tile.blp
```diff
@@ -0,0 +1,139 @@
...
+    Box {
+      hexpand: true;
+      halign: end;
+      margin-end: 8;
+      valign: center;
+      spacing: 4;
+
+      Image {
+        icon-name: "circle-crossed-symbolic";
+        tooltip-text: _("This extension is not comptabile with your current version of GNOME.");
```
AsciiWolf:
```suggestion
        tooltip-text: _("This extension is not compatible with your current version of GNOME.");
```

Typo.

AsciiWolf:
Would it be possible to disable this? E.g. for non-GNOME setups? And do we really want to change Bazaar to a general purpose software store (like GNOME Software / KDE Discover)? I am a little bit puzzled about this, to be honest.

kolunmi:
yes we should disable this in non-gnome sessions

AlexanderVanhee:
> Would it be possible to disable this? E.g. for non-GNOME setups? And do we really want to change Bazaar to a general purpose software store (like GNOME Software / KDE Discover)? I am a little bit puzzled about this, to be honest.

Its already disabled for non GNOME environments

PR #1468, opened 2026-05-09
Update Simplified Chinese translation

Thundercup:
Update Simplified Chinese translation for new version


kolunmi:
Thanks so much! :cn:

PR #1469, opened 2026-05-09
Animated Transact Icon

kolunmi:
https://github.com/user-attachments/assets/2e1a1039-0ecb-49d4-90dd-9c9fdf3f59ec


AlexanderVanhee:
I would make the path width and shrunken down icon size constant between states, like 0.6 for the icon-scale and 3.0 for the path-width.

We will also need support for fallback icons, as otherwise the animation would not be visible.

kolunmi:
binding onto the wdgtrenderer also slows things down

kolunmi:
https://github.com/bazaar-org/bazaar/pull/1469/changes/ae70a0eb9792bb47726df81aef667fa0d64d24a3 seems to fix the slowdowns

PR #1477, opened 2026-05-11
Allow app cache deletion

AlexanderVanhee:
Also switched out our custom recursive folder size walking function with a Gio wrapper.

Screenshot From 2026-05-11 13-57-32


kolunmi:
ehh sure why not, thanks!

PR #1484, opened 2026-05-13
Fix install button not having the suggested action

AlexanderVanhee:
For some reason, get_install_btn_state has the side effect of adding or removing the "suggested-action" class from the install button. However, because the fallback didn't call a function, it also didn't manage the class correctly. Ideally, there wouldn't be any side effects, but managing styles through bindings is not possible so we have to piggy back.


kolunmi:
thanks!

PR #1485, opened 2026-05-13
Fix transaction level error reporting

AlexanderVanhee:
Most errors are specific to a job, but there are also transaction level errors, most easily seen when not entering a password in a polkit prompt. These were seemingly ignored before, as they were returned as an error hashmap. Now this hashmap gets properly read and set as an error, and the transaction gets also marked as cancelled.

This fixes the issue where the bundle install dialog reports the app as installed when the user did not provide their password, even though the installation did not actually complete.


kolunmi:
LGTM

PR #1492, opened 2026-05-14
Ensure runtime remote before bundle install

AlexanderVanhee:
Bundle installs would fail if the runtime remote was not already added beforehand. Now it gets added automatically if need be (like how already said so in the UI).

`flatpak_remote_new_from_file` does not set GPG verification correctly for some reason... So i had to do it manually. 

The URL below provides a bundle with `gnome-nightly` for the runtimes remote:
https://gitlab.gnome.org/GNOME/gnome-text-editor/-/jobs/6530959/artifacts/download?file_type=archive

Make sure you don't have `gnome-nightly` already when testing.

### DIFF
--- a/src/bz-bundle-install-dialog.blp
+++ b/src/bz-bundle-install-dialog.blp
@@ -178,11 +178,12 @@ template $BzBundleInstallDialog: Adw.BreakpointBin {
 
                   Label {
                     label: _(
-                      "Installing this app requires adding a new software source. Other apps from this source will show up in Bazaar.\n\nOnly add this source if you're sure you trust it."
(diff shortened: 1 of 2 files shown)


kolunmi:
thanks! please rename `ensure_runtime_remote` -> `fiber_ensure_runtime_remote` since it has a dex_await in it and you are just calling it normally from a fiber, and make sure it is forward declared at the top of the file.

Also `local_error` should be the first declared variable in the scope

AlexanderVanhee:
Should be better now

kolunmi:
looks good

PR #1493, opened 2026-05-14
Minor permission rating changes

AlexanderVanhee:
Permissions to subfolders are now hidden if access has been given to the full home or root.

 The camera and individual filesystem danger levels are now also downgraded in severity if there is no network access.

part of #1454


kolunmi:
thanks!

PR #1494, opened 2026-05-14
Stop the transact fiber from referencing a window

AlexanderVanhee:
This fixes the issue where the window would open at a weird resolution when a transaction was running in the background.

Opening a new window mirrors the resolution of the primary window if it exists. In this case the window was closed but still referenced, causing the new window resolution to be set to 0x0.

Closes #1465


kolunmi:
thanks

PR #1495, opened 2026-05-14
Adress "installed" refresh worker race condition

AlexanderVanhee:
There seems to a rc where the refresh worker could replace an entry in cache after the last replace_entry in BzApplication, causing the entry to be marked as not installed.

My fix is to find the installed entries and overwrite them again in enumerate_disk_entries_fiber, so after when the cache manager is done.

This seemingly fixes the issues with runtimes not being marked as installed, or apps from other remotes that are not Flathub something not wanting to uninstall.

You might still want to look into a better solution tough.


kolunmi:
closing in favor of #1496

PR #1497, opened 2026-05-14

AsciiWolf:
No strings from _src/bz-bundle-install-dialog.blp_ were translated.


kolunmi:
Good catch!! Thanks

PR #1499, opened 2026-05-14

theMimolet:
Hiii,

I updated the strings for the new bundle feature + refreshed some :D

It's not ready yet as there's [a string](https://github.com/bazaar-org/bazaar/blob/main/src/bz-bundle-install-dialog.blp#L126) that require to be treated like the other install buttons strings (otherwise it makes it weird in french) and it still needs a few tweaks here and there.

Hopefully I'll wake up early enough to open the PR before the update ><

Anyway, love you all, 
Have a lovely night <3


theMimolet:
Here's the French translation ! 

Thank you for your awesome work <3

kolunmi:
Thank you for your awesome work as well!

PR #1500, opened 2026-05-15
Handle no runtime entry in the UI

AlexanderVanhee:
Also fixes some errors that would show up for apps where the ui-entry was not from Flathub.


kolunmi:
thanks!

PR #1503, opened 2026-05-16
Remove global separator opacity override

AlexanderVanhee:
This was needed in some previous design of the update card, don't ask me why I didn't scope it properly


kolunmi:
Thanks!

PR #1504, opened 2026-05-16
Fix global search when in subpages

AlexanderVanhee:
When searching from the full view or whatever, only the first keystroke would appear in the search bar, and the page would not switch.


kolunmi:
Thanks!

PR #1506, opened 2026-05-16
Add cancel button to rich app tile

AlexanderVanhee:
Adding a cancel button allows us to disable to confirmation dialog, saving the users an extra click.

image


theMimolet:
Would it be pertinent to replace "Get" by a download icon ? 
(absolutely not biased by the fact it's "Installer" in French and that we barely see the app description)

kolunmi:
Thanks!

PR #1516, opened 2026-05-17
Update devcontainer configuration and Dockerfile dependencies

pan0xc:
1. Add the necessary dependencies to the Dockerfile for successful compilation.
2. The absence of the dbus component within the container causes Bazaar to crash. Exposing the host machine's dbus component within the container allows for its use.

Local testing is working fine; the only issue so far is that the flatpak app is not working.
image


kolunmi:
Thanks!!

PR #1525, opened 2026-05-19
Stop putting name and icon names of links in entries

AlexanderVanhee:
This of course causes issues when changing languages. As its still cached in the previous language.

Closes #1522

image


kolunmi:
Thanks!!

PR #1526, opened 2026-05-19
Allow translators to use download icon for "Get" button

AlexanderVanhee:
Is useful if there isn't a short enough translation available in that language.

image


kolunmi:
thx

PR #1527, opened 2026-05-19
Hide header bar title for apps page if regular title is visible

AlexanderVanhee:
Copies that animation thingy from the full view

image


kolunmi:
thanks!! :t-rex:

PR #1530, opened 2026-05-20
chore: actually link to a new website

Lumaeris:
Randomly noticed that Bazaar as of 0.8.0 still links to a repo when a proper website exists


kolunmi:
thanks!!

PR #1533, opened 2026-05-22
Fetch updates in async after cache load

AlexanderVanhee:
Updates to Bazaar where not fetched after cache load ever since I removed that call to improve the loading speed. What I actually should have done is to fire and forget.

Closes #1532


kolunmi:
closing in favor of https://github.com/bazaar-org/bazaar/commit/1bf0e05c511ee3fea2d14f3ec7179e140e3ff686

Issue #1442, opened 2026-04-24
what is bge library?


Hasanhasan323:
HI,
Currently you add a new project bazaar, bge.
Could you explain it more? What is it? what does do?

kolunmi:
bge stands for Bazaar GTK Extensions, it's a utility library that provides new widget implementations, animation helpers, etc

Hasanhasan323:
Can this library be used by programmes other than  bazaar?

kolunmi:
yep!

Hasanhasan323:
And what is /usr/bin/bge-demo?
What does this do?

PR #1539, opened 2026-05-23
bge: fix shared library versioning

ndpm13:
I've encountered an issue while packaging Bazaar, caused by the bge library embedding the version in the library's base name.

This patch fixes the issue for me, resulting in the conventional `libbge.so` → `libbge.so.0` → `libbge.so.0.x.x` symlink chain and a stable `bge.pc`.

This change is only required if bge is meant to be a public library, which is the case from what I understood according to #1442.


kolunmi:
thank you! appreciate it

PR #1540, opened 2026-05-23
chore(po): Update Brazilian Portuguese translation

koitorin:
Should have done this last week but better late than never!!!


kolunmi:
Thanks! :brazil:

PR #1541, opened 2026-05-23
Display a couple of add-ons on the full view

AlexanderVanhee:
Makes it so a maximum of 3 add-ons are shown on the full view page, with a button at the bottom to show all add-ons if there are more. This should speed up add-on management in most cases, as most apps only have 1 add-on.

Also adds labels to some sections, similar to what is seen on Flathub. This should make the logical sections easier to parse.

Screenshot From 2026-05-23 20-10-26
Screenshot From 2026-05-23 19-44-13


kolunmi:
Thanks

PR #1542, opened 2026-05-24
Fix category-less state of search page

AlexanderVanhee:
The status page was above the other elements for some reason.


kolunmi:
Thanks

PR #1543, opened 2026-05-24
Make bundle install dialog more adaptive

AlexanderVanhee:
It's best to hide whitespace changes for this one

image


kolunmi:
Thanks

PR #1544, opened 2026-05-24
Rework Flathub offline page

AlexanderVanhee:
Previously there were 2 offline pages, both with different actions. The one visible with "has-connection-error" was never visible as the property goes unused it in the newer versions of the app.

So I merged the 2 pages and removed the unused property.

image


kolunmi:
Thanks

PR #1545, opened 2026-05-24
Disable error due to bad bus policy

AlexanderVanhee:
org.kde.Platform/x86_64/6.11-beta seems to have an invalid key under [Session Bus Policy] which causes an error at every startup.


kolunmi:
thanks

PR #1548, opened 2026-05-24
Select the second screenshot at the start again

AlexanderVanhee:
This functionality was lost when moving to the bge-carousel. I also believe I fixed the bug where it seemingly randomly wasn't set to the second one. I believe the reason was that we were checking n-items, but that won't work if the number of screenshots stays the same.


kolunmi:
thanks

PR #1549, opened 2026-05-25
Bring back the danger style for desktop only apps on mobile widths

AlexanderVanhee:
I was asked to remove the icon change based on width of the full view, but accidentally also removed the style changing aswell


kolunmi:
thanks

PR #1556, opened 2026-05-26
Remove pride flag rotation toggle

AlexanderVanhee:
Changed it so the install controls always use the horizontal variants, as flags with many colours look better on the long pill. And vertical for the transact icon.


kolunmi:
thanks

PR #1558, opened 2026-05-26
Create and make use of a cache for entry groups.

AlexanderVanhee:
We can save a lot of time by not having to deserialize all the entries at startup just to be able to create the entry groups.
All functionality that I tested still seems to work fine.

On my PC, the time from the init timer creation until the end of the init fiber is now around 825 ms. Ideally, I want to reduce it even further by looking for other optimizations, such as parallelizing the deserialization of the Flathub and entry group caches, or parallelizing or deferring some other tasks that happen during initialization. All for the ultimate goal of nothing having a loading timer visible to the user.


kolunmi:
Looks good, thanks!

PR #1562, opened 2026-05-27

AsciiWolf:
Also remove the permission setting fuzzy strings that are no longer useful.


kolunmi:
thank you

PR #1563, opened 2026-05-27
Rework the share list

AlexanderVanhee:
Adds a dual column mode to the links, hides the scheme for the URL and changes the "external" button into an icon.

image


kolunmi:
thanks

PR #1571, opened 2026-05-29

Micro856:
Small update to Polish translation.

P.S. I finally reset the commits on my repo. Of course had to use git cli for it,  couldn't use web ui.


kolunmi:
thanks

PR #1575, opened 2026-05-31
Make the installed donation button have its link as its tooltip

AlexanderVanhee:
This makes it more obvious that the button will open a URL, it also follows the pattern of the other donation button in the full view.


kolunmi:
Thanks

PR #1576, opened 2026-05-31
Add manifest link sometimes so the columns stay filled

AlexanderVanhee:
The link to the manifest is not that important to have, but it's still better than showing an empty space

image


IsabelleRoux:
Hi @AlexanderVanhee, rather than adding a link to the manifest that is not very useful in order to avoid an empty space, why not simply remove the Donate button, which is redundant with the Support button (it does the same thing)?

AlexanderVanhee:
We want to make the donation link as easily accessible as possible, and because the donation link is in the list in GS, I believe we should have it there as well.

kolunmi:
Thank you

PR #1577, opened 2026-05-31
Fix the fallback for the auth avatars

AlexanderVanhee:
Screenshot From 2026-05-31 14-30-59
Screenshot From 2026-05-31 14-24-45


kolunmi:
Thanks

PR #1578, opened 2026-05-31
flatpak-instance: detect user-installed flathub remote

lechl1:
## Summary
- `check_has_flathub_fiber` skipped user remotes under `SANDBOXED_LIBFLATPAK`, so a user-installed `flathub` was invisible to Bazaar — the "empty page" / no-categories state stuck even when Flathub was actually configured.
- `ensure_flathub_fiber` had a parallel sandbox branch that returned success without doing anything when the system Flathub was missing, silently no-op'ing the "Set Up Flathub" action and re-prompting on every launch.
- Drop both `#ifdef SANDBOXED_LIBFLATPAK` guards so the unified user-then-system path is used in both builds.

## Test plan
- [ ] `SANDBOXED_LIBFLATPAK` build with only a *user* Flathub remote: launch and confirm categories populate (no empty-page state).
- [ ] `SANDBOXED_LIBFLATPAK` build with no Flathub anywhere: clicking "Set Up Flathub" in the prompt actually adds Flathub (no silent no-op, prompt does not return on next launch).
- [ ] Non-`SANDBOXED_LIBFLATPAK` build: behavior unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/bz-flatpak-instance.c
+++ b/src/bz-flatpak-instance.c
@@ -842,8 +842,6 @@ check_has_flathub_fiber (CheckHasFlathubData *data)
       n_system_remotes = system_remotes->len;
     }
 
-// Downloading from user remotes in the Flatpak is unsupported.
-#ifndef SANDBOXED_LIBFLATPAK
   if (self->user != NULL)
     {
       user_remotes = flatpak_installation_list_remotes (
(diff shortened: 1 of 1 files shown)


AlexanderVanhee:
This won't work for apps that use "extra_data" like google Chrome

PR #1581, opened 2026-06-01
Add Network subcategories

AlexanderVanhee:
Probably the most usefull one of the ones we don't already have subcategories for.


kolunmi:
Thank you

PR #1593, opened 2026-06-06
Add Mobile Search Filter & rework its "all apps page"

AlexanderVanhee:
Whilst we can't add a toggle in preferences to filter out not mobile apps globally due to the UI issue, we can add it to search without a problem.

Instead of adding a new bool property to entry group, I opted to just set the mobile category flag (that already existed) and consume that instead.

This also means that we can trivially convert its "all apps page" to actually show all mobile apps instead of just showing the collection overflow, like we do with "popular" and such.

Screenshot From 2026-06-06 20-47-40

Closes #1591

### DIFF
--- a/src/bz-appstream-parser.c
+++ b/src/bz-appstream-parser.c
@@ -609,6 +609,9 @@ bz_appstream_parser_populate_entry (BzEntry     *entry,
         }
     }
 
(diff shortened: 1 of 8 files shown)


kolunmi:
thanks

PR #1596, opened 2026-06-07
Fix infinite fetching of user data folder size

AlexanderVanhee:
This infinite loop caused CPU usage to spike, and probably disk access too, especially for installed apps with large user data or cache directories, like web browsers.

Fixes #1594 

This probably also fixes this issue:
Fixes #1507


kolunmi:
thanks

PR #1599, opened 2026-06-07
Improve full view partially empty state

AlexanderVanhee:
Improves the handling of apps without or with partial metainfo files

Screenshot From 2026-06-07 15-12-00


kolunmi:
thanks, but please format the code in your prs from now on

PR #1600, opened 2026-06-07
Offload the cache enum disk read to IO scheduler

AlexanderVanhee:
This fixes a very visible hitch in the middle of loading on a non cached app launch (around the 8 to 9 second mark).


kolunmi:
thanks

PR #1607, opened 2026-06-09
Fix double focus on the rich card

AlexanderVanhee:
You would need to press tab twice before going to the next element.


kolunmi:
thanks

PR #1608, opened 2026-06-09
Refactor search bars + improve search accessibility

zeldaAhmed:
Draft because theres an issue with focusing the search bar causing it to somehow draw focus rings around both the end-child and the search bar when focusing into the search bar, ~~and I need to run clang-format again.~~

This MR fixes quite a few of the issues I found with both keyboard navigation + screen readers I found within Bazaar
* The search bars did not read out their placeholder text when focused, and not containing any text
* Ctrl+F focused the last widget and not the search bar when the search page is presented
* Bazaar did not read out the number of search results to the screen reader.

Before:

https://github.com/user-attachments/assets/9d43a727-f46d-4d65-b111-e0a4318b3891

After:

https://github.com/user-attachments/assets/6c7bebea-73fc-4230-b863-19bf80e64285


AlexanderVanhee:
I fixed the double focus thing, but I can't push to this, so here is the patch file.

[fix-search-focus.patch](https://github.com/user-attachments/files/28911894/fix-search-focus.patch)

kolunmi:
Thanks!!! :rat:

PR #1613, opened 2026-06-10
Fix cache version only being created at the second start

AlexanderVanhee:
The old logic would cause the app to remove the cache at second start, causing a second long load time and the donation banner to show, as the cache version file would not yet exist.


kolunmi:
Thanks!

PR #1616, opened 2026-06-11
Disable the focus on the rich app tile instead of the list item

AlexanderVanhee:
Our previous approach stopped the arrow navigation from working


kolunmi:
Thanks

PR #1618, opened 2026-06-13
Increase spacing in full view top pills to match others

AlexanderVanhee:
Always thought the spacing there was a bit too small


kolunmi:
thanks

PR #1621, opened 2026-06-13
Enumerate disk entries in batches

AlexanderVanhee:
If you opened the full view while it was almost done refreshing the entries, specifically when the disk entries were being deserialized, it would display a broken page for a second or so. This happened because the entry it needed for the UI was placed at the end of the queue, behind all the entries from the enumeration.

[Screencast From 2026-06-13 17-05-22.webm](https://github.com/user-attachments/assets/10c94b60-edb4-48f9-a68c-97739ba62cbc)

### DIFF
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -23,6 +23,8 @@
 
 #define MAX_IDS_PER_BLOCKLIST 2048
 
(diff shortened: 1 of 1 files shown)


kolunmi:
thanks

PR #1622, opened 2026-06-13
Better handle slow entry decaches in UI

AlexanderVanhee:
Added a widgetless "loading" screen for the full view in the rare circumstance where the entry is not loaded instantly, masking the broken state the full view shows when no entry is available.

Also prevents both the unloaded screenshot and the "no image" icon from being shown next to each other when an entry has not yet been deserialized.


kolunmi:
thanks

PR #1626, opened 2026-06-14
Focus the first install or remove button when pressing enter in search bar

AlexanderVanhee:
The idea is that people who want to press Enter twice to install an app can still do that, while not annoying people with muscle memory by triggering a dialog they didn't intend.

Closes #1595


kolunmi:
thanks

PR #1632, opened 2026-06-17
Make featured carousel accessible

AlexanderVanhee:
Makes the main "app" part of the widget focusable and activatable. Focusing now stops the auto rotation. Also, it now announces the title and description of the app when navigating through the carousel to the screen reader.


kolunmi:
thanks

PR #1633, opened 2026-06-17
Make important full view text selectable

AlexanderVanhee:
This is needed for accessibility, it also includes a fix for the selected state of the description renderer, so it selects text like a normal label would do.


kolunmi:
thank you

PR #1638, opened 2026-06-19
chore(po): Update French translations

theMimolet:
Here you go :D 

(i had such a big day, I'm so tired ;-;)


kolunmi:
thank you!! :fr:  *Pats

PR #1639, opened 2026-06-19
Trigger show-group action from app tiles directly

AlexanderVanhee:
This reduces boilerplate, and makes the tiles more easily reusable.


kolunmi:
Thanks

PR #1641, opened 2026-06-20

Lumaeris:
spooky permission granting all permissions the app wants: :ghost:


kolunmi:
Thank you!

PR #1655, opened 2026-06-22
Rework Curated System

AlexanderVanhee:
This PR reworks the Curated page, improving accessibility, making it fit better with the rest of the app, and providing more options for vendors.

It replaces the single section widget featuring an image with overlaid text (which was often difficult to read) and app tiles that were heavily styled with CSS, with a set of widgets, some of which are repurposed from other parts of the app.

It also adds articles, allowing vendors to display news in a place where users are more likely to see it, or to publish articles about apps.

An example yaml file can be found in docs/example.yaml

image
image
image
image


kolunmi:
This is incredible!

renner0e:
Pretty awesome, would be nice to get a heads-up before this is merged, as it breaks compatibility with the old format. Thanks!

kolunmi:
Thanks!


https://github.com/user-attachments/assets/60b50928-1f2d-4414-aef8-2076e2e9ea99

PR #1662, opened 2026-06-24
Change screenshot opening animation

AlexanderVanhee:
https://github.com/user-attachments/assets/005a1c15-4010-47e2-adc6-268ff85723f2

### DIFF
--- a/src/bz-article.c
+++ b/src/bz-article.c
@@ -37,6 +37,7 @@
 #include "bz-screenshot-page.h"
 #include "bz-state-info.h"
 #include "bz-util.h"
+#include "bz-window.h"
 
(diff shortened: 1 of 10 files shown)


kolunmi:
Looks great! Could you do a spring animation instead of timed so the movement takes account of the distance it has to cover? Also perhaps make the animation curve a bit less harsh towards the end

kolunmi:
also the back button on my mouse no longer works to exit the screenshot mode :(

kolunmi:
merged with cli: https://github.com/bazaar-org/bazaar/commit/93cf1bdb30b49ec68b71585f2b088bc4165fff9c

PR #1664, opened 2026-06-24
Snap to selection on map

AlexanderVanhee:
This fixes an issue where when you returned from the full view or whatever, it would show the carousel stopped somewhere in the middle between 2 featured tiles.


kolunmi:
superseded by https://github.com/bazaar-org/bazaar/pull/1674

PR #1666, opened 2026-06-25
Dont set fallback acessiblity label for install button

AlexanderVanhee:
It does nothing and messes up translations.

Fixes #1653


kolunmi:
thank you

PR #1667, opened 2026-06-25
Show just the version number in the update row if it hasn't changed

AlexanderVanhee:
As it would look weird to show "v1->v1", I initially decided to just not show anything in the subtitle in that scenario. But it's probably a better idea to still just show the version, like in GS.

Screenshot From 2026-06-25 12-44-05

Fixes #1654


ThiloteE:
Maybe a little sentence like "Oldversion -> "Update is available, but version number remains unchanged" would be ok? Maybe too large for UI? I can't think of something shorter.

PR #1668, opened 2026-06-25
Move safety dialog carousel scroll to map

AlexanderVanhee:
This fixes a race condition where the sandbox escape page would sometimes show while it shouldn't. The GTK docs recommend starting animations on map.


kolunmi:
thanks

PR #1671, opened 2026-06-26
Stop auto syncing on metered connection at startup (or after 24h)

AlexanderVanhee:
The update fetching still happens automatically, the banner also re-appears after daily background refresh fails.

Screenshot From 2026-06-26 18-05-59


kolunmi:
Thanks

PR #1677, opened 2026-06-28
Move auth state setup to init fiber

AlexanderVanhee:
It blocked the main thread for about 25ms, by far the worst step of `init_service_struct`.


kolunmi:
Thanks

PR #1680, opened 2026-06-29
Cache article thumbnails

AlexanderVanhee:
This should hopefully reduce the pop in of the article thumbnails, important if we want the option to show the curated page by default.

I don't think it makes much sense to do it for the article images itself, as revisits are probably rare, and so would only take unneeded space on disk.


kolunmi:
wait the nitpick wasn't relevant to this pr, mb

kolunmi:
thanks!

PR #1685, opened 2026-06-30
Show a warning if using the Flatpak on Ubuntu

AlexanderVanhee:
I made sure to only check it the first time, so users who manually fixed it can still use the app without constant warnings. This also makes it so the check isn't ran every time on all distros, saving a precious 20ms.


kolunmi:
awesome, but could you use the libdex subprocess spawning function since you are in a fiber? example: https://github.com/bazaar-org/bazaar/blob/bd1a49cc89cddf692eff9d9efb4b3e41cfdadb1c/src/bz-hooks.c#L408-L433

renner0e:
CC: https://github.com/bazaar-org/bazaar/issues/583

kolunmi:
thanks

PR #1687, opened 2026-07-01
bge: MarkdownRender: add support for tables

AlexanderVanhee:
Per column alignment is not supported.

image

image


kolunmi:
Thanks!

PR #1688, opened 2026-07-01
Recalculate targets position on each frame

AlexanderVanhee:
This fixes any weirdness when closing the apps page or moving the carousel when the closing animation is still playing.

I'm pretty sure it even fixes the jump due to the screenshot being raised on hover.


kolunmi:
Thanks!

PR #1691, opened 2026-07-02
Don't skip initial refresh on metered connection when there is no cache

AlexanderVanhee:
You seem to also have accidentally removed the metered connection banner in this commit: https://github.com/bazaar-org/bazaar/tree/93cf1bdb30b49ec68b71585f2b088bc4165fff9c


kolunmi:
Thanks!

PR #1700, opened 2026-07-03
chore(po): update Indonesian translation for 0.9.0

KiKaraage:
Here's some beef from my fam's last month cooking~

image


kolunmi:
yummy!

PR #1701, opened 2026-07-03
Add reduced motion alternative animation for the screenshot page

AlexanderVanhee:
It's intentionally a bit slower than other page changes to make the transition from dark mode window to a bright white screenshot easier on the eyes.

[Screencast From 2026-07-03 17-55-04.webm](https://github.com/user-attachments/assets/c8f7f84d-fb88-480a-9fe4-85b1db1232ed)


kolunmi:
Thanks!!

PR #1703, opened 2026-07-04
Add article custom tile hook

AlexanderVanhee:
Also removed the "preferred-width" property of the app-tile to simplify it, as it went unused anyways.

image

image

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -396,7 +396,8 @@ Here is how they did it:
 
 ## Article markdown
 
-Curated articles support standard markdown, rendered with GTK widgets (so its not a webview). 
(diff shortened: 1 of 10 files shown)


AsciiWolf:
Unrelated question: Is the `Install via ...` dialog translatable?

kolunmi:
> Unrelated question: Is the `Install via ...` dialog translatable?

@AsciiWolf It's a hook dialog string, so this is just an example. It would need to be translated on the distro's side of things in the yaml

kolunmi:
@AlexanderVanhee thanks! I tweaked a few things, but it looks good now. I appreciate the pr

PR #1705, opened 2026-07-05
Add "similar apps" section to full view

AlexanderVanhee:
This PR adds a new section to the full view page showcasing similar apps to the one the user is currently viewing. Making it easier to discover alternative or related apps.

I have added a daily GitHub action to the site that runs a python script. The script first downloads and parses the Flathub appstream bundle, and then tries to find similar apps for each app found via TF-IDF and cosine similarity via solely the apps metadata. It also avoids other apps by the same dev, as the other section already shows that.

image


kolunmi:
sick

PR #1706, opened 2026-07-05
Use new features from Dex 1.2

kolunmi:
Begin using DexLimiter (from Dex 1.2) in places instead of the hacky macros from bz-util.h

Will need to do some testing to figure out what the best values are


kolunmi:
see: https://gnome.pages.gitlab.gnome.org/libdex/libdex-1/class.Limiter.html

kolunmi:
@chergert if you have any suggestions as well based on the diff, please do let me know. Does it make sense for the max concurrent io tasks to be defined in relation to `g_get_num_processors`? Thanks so much for creating libdex <3

chergert:
Generally I would pin to core count. Since `g_get_num_processors()` doesn't guarantee hyperthreading or not, the only real way to get that is to look at the CPU topology. And that isn't even standardized across architectures in `/proc/cpuinfo` which makes that a rather annoying thing.

So I tend to just do what you've done there, n_cpu/2.

chergert:
I should also briefly mention, that if this is wrapping something like `glycin`, you'll probably find it is creating yet another thread pool per process (n_cpu in size) from radon or whatever the underlying codecs are using, so the contention can happen externally to you no matter what you are limiting here.

I'm not a fan of the glycin design in this standpoint, but that ship has sailed, at least as far as upstream GNOME is concerned.

fossdd:
was it intentional to use a beta libdex for a stable bazaar release?

AlexanderVanhee:
Yes, I recommend you either skip this update or statically link libdex for now.

fossdd:
Alright, makes sense. Thanks for clarifying.

AlexanderVanhee:
Sorry about the inconvenience.

PR #1712, opened 2026-07-06
Window: Fix avatar menu button being visible on load

AlexanderVanhee:
Since I moved the auth init to the fiber, auth-state would be null for a little bit. Causing the MenuButton visibility expression to fail and return true.


kolunmi:
thanks

PR #1713, opened 2026-07-06

theMimolet:
- Added the new translations
- Redid some of the old ones (I took some inspiration from GS)


kolunmi:
thanks!

PR #1715, opened 2026-07-07
Don't have a timeout for clearing search text

AlexanderVanhee:
Only slows things down for no reason


kolunmi:
I would prefer if you just called `update_filter` directly instead of passing 0 as an interval argument to `timeout_add`

AlexanderVanhee:
ok

kolunmi:
Thanks!!!

PR #1716, opened 2026-07-07
Rework leftover user data page

AlexanderVanhee:
- Show data we don't have an entry-group for by creating a fake group
- Add a folder button similar to the one in the size dialog
- Stop Bazaar from showing up the list
- Fix code style issues

image


kolunmi:
thanks

PR #1717, opened 2026-07-08
Add placeholder size for non loaded screenshots

AlexanderVanhee:
This stops ugly layout shifts in the screenshot carousel, especially noticeable on slower internet connections. It's also especially important for us as we start from the second screenshot.

It's possible that some remotes don't provide width and height in their appstream, but in that scenario it should just fallback to the old behavior.

(I added the red boxes and loading delay for illustrative purposes)
[Screencast From 2026-07-08 10-59-42.webm](https://github.com/user-attachments/assets/48505afc-f48e-414c-8d9b-1055e44a3e51)


kolunmi:
Thanks, looks great

PR #1721, opened 2026-07-09
Open app page in background when opening add-on via appstream

AlexanderVanhee:
Also fixes issue where the add-on dialog could spawn on top of another dialog.

I made to choice to bind to the property in the window so it stays clear what action_show_group does exactly.

image

closes #1709

### DIFF
--- a/src/bz-addons-dialog.c
+++ b/src/bz-addons-dialog.c
@@ -325,6 +325,16 @@ bz_addons_dialog_new_single (BzEntryGroup *group)
   return ADW_DIALOG (self);
 }
 
+BzEntry *
+bz_addons_dialog_get_parent_entry (BzAddonsDialog *self)
+{
+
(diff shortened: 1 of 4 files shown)


src/bz-addons-dialog.c
```diff
@@ -325,6 +325,16 @@ bz_addons_dialog_new_single (BzEntryGroup *group)
   return ADW_DIALOG (self);
 }
 
+BzEntry *
+bz_addons_dialog_get_parent_entry (BzAddonsDialog *self)
+{
+
```
kolunmi, line 332: Type check

kolunmi:
Thanks! Left a comment, also not a fan of calling a property notifier explicitly but window.c is and always was a disaster anyway

AlexanderVanhee:
Adding a signal would be an ass ton of boilerplate though.

kolunmi:
Thanks!!!

PR #1726, opened 2026-07-10
Allow gradients in curated section banner bg

AlexanderVanhee:
I thought you could do this, but I used the wrong css propery :(


kolunmi:
Thanks

PR #1732, opened 2026-07-11
Add daemon process

AlexanderVanhee:
The daemon process is a wrapper around the actual Bazaar application, it's responsibilities are managing the application, forwarding launches with parameters (except --no-window) and implementing the search provider (also auto-updates in the future).

The search provider part consumes a ~2mb file with the data needed for the search engine, its generated after the entry-group cache.

The daemon does not link to gobject library to keep the memory footprint as low as possible, around 8 mb with a further 12mb for Flatpak stuff. I believe the recent improvements with launch speeds it makes sense to not keep the main application alive, especially when RAM is so expensive nowadays...

You will have to reinstall the Flatpak app to test everything out properly


kolunmi:
https://github.com/user-attachments/assets/6e176822-b15c-4bb0-8b1c-a1347abda193

PR #1733, opened 2026-07-12
Implement method launch search properly

AlexanderVanhee:
This action was seemingly also broken in the old implementation.


kolunmi:
nice, we should probably wrap malloc in a static function that bails if it returns null tho

PR #1737, opened 2026-07-12
fix(ci): workaround broken caching mechanism upstream

renner0e:
The short explanation is that we either get a stale cache or never
restore a cache without this. We are hitting the former.

The bad part about this, is that cache is incremental only, but it would
get pruned when it reaches 10G.

I have no idea why bazaar itself is never cached, only the deps for some
reason. The build does not take long anyway, so this change as a whole
is purely for the love of the game.


renner0e:
search for "cache hit" here in this log
https://github.com/renner0e/bazaar/actions/runs/29204656419/job/86681881929

kolunmi:
Thanks!

PR #1738, opened 2026-07-13
fix(po): Remove src/bz-gnome-shell-search-provider.c

AsciiWolf:
It was removed in 3dd1497f42d081dd57ebdff1ccf085836889e687.


kolunmi:
Thanks!!!

Issue #805, opened 2025-12-13
feat: Automatic/background update installation


linfindel:
Apologies if this is a duplicate, I can't believe that no-one has suggested this before but I can't find a matching issue (other than https://github.com/kolunmi/bazaar/issues/547 which is more about controlling automatic checking for updates)

Presently, Bazaar checks for updates in the background and notifies the user upon opening Bazaar. It would be nice if it could (optionally) do the update in the background as well :)

This feature is included in the design here: https://github.com/kolunmi/bazaar/issues/547

PR #1739, opened 2026-07-13
Add an auto updater

AlexanderVanhee:
The auto updater is disabled by default, to not interfere with the auto updaters of ublue or some other app store. Even if running 2 auto updaters probably won't cause any actual issues.

The logic for when to update is the same as in GNOME Software, except it does not check for GameMode and the random hour offset is only set once.

I chose to not re use any backend code from the Flatpak instance, as the update logic is fairly small, dissimilar and does also not need regression testing whenever we change that code.

image


linfindel:
Would close #805

kolunmi:
looks good thenks :)

PR #1740, opened 2026-07-13
Add timeout for startup refreshes

AlexanderVanhee:
As the application gets closed now after the last window gets closed, it's unwise to refresh after every application start. As it uses both CPU and bandwidth for little reason.


kolunmi:
thanks!

PR #1744, opened 2026-07-14
Exit daemon on idle when background update is off

AlexanderVanhee:
This PR fixes the dbus activation of the search provider, this means if no auto updates are requested by the user. We can just close the daemon after a couple of seconds and reopen only it when the user starts searching again.


kolunmi:
pog thanks

PR #1745, opened 2026-07-16
Window: expose "app.preview-metainfo" action in main global menu

kolunmi:
https://github.com/user-attachments/assets/3f956451-f0db-4253-9136-94f0000db5d5

### DIFF
--- a/src/bz-window.blp
+++ b/src/bz-window.blp
@@ -327,6 +327,12 @@ menu primary_menu {
   }
   section {
     item {
+      label: _("Preview Meta_info File");
+      action: "app.preview-metainfo";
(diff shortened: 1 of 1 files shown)


AsciiWolf:
To be honest, I am not sure if it is a good idea to add developer-focused features directly in the Bazaar main menu.

vmkspv:
> To be honest, I am not sure if it is a good idea to add developer-focused features directly in the Bazaar main menu.

This may be useful if the user is logged in with Flathub and is a developer. Just need an extra check.

AlexanderVanhee:
Still needed?

PR #1759, opened 2026-07-22
Update metainfo preview system

AlexanderVanhee:
Swaps the launch parameter out with a metainfo file scheme handler, where the user first opens the metainfo file (like you would with a .flatpak), then gets to option to choose an icon via a dialog.

Also reworks the metainfo preview page to contain other apps with high quality metadata. So the user more easily compare their own work to other high quality entries.

Screenshot From 2026-07-22 15-03-38

image


kolunmi:
Thanks

PR #1760, opened 2026-07-22
Improve desktop entry searchability

vmkspv:
Adds a GenericName and clearer Keywords to the desktop entry so Bazaar shows up when searching for terms like `Store` or `Package` in app launchers. This should also help with localized searches via GenericName (closes #1585).


kolunmi:
Thank you!!

PR #1761, opened 2026-07-22
Handle EOL-rebasing

AlexanderVanhee:
Handles EOL rebasing as a kind of special case of updates, like GS seems to do. Flatpak seems to also return apps that need to be rebased as updatable.

The core part is just a different Flatpak API call, but we also need some plumbing to handle the removal of the old app from the installed app list.

You can use `flatpak install dev.bragefuglseth.Keypunch` to check it out in action, just make sure to say no to the prompt were it asks to install the new version instead.


kolunmi:
Thank you!

PR #1763, opened 2026-07-24

Micro856:
Updates Polish translation and changes some strings to better match GNOME Polish Team way of translating, because of plans to [move Bazaar translation over to Damned Lies.](https://gitlab.gnome.org/Teams/Translation/Coordination/-/work_items/534).


kolunmi:
Thanks!! :poland:

PR #1766, opened 2026-07-27
Add translators notes & remove unneeded strings

AlexanderVanhee:
Most tiles in Adw.ViewStack never get shown to the user, as they only get used when using a view switcher, so it makes sense to remove them to spare the translators some work.


kolunmi:
Thank you

PR #1767, opened 2026-07-28
User Data Fixes

AlexanderVanhee:
Makes it so the empty page gets properly shown when the last entry gets removed.

Also makes it so only 1 toast can exist at a time, updating the count when every time the user removes an entry.

Screenshot From 2026-07-28 22-02-49
Screenshot From 2026-07-28 22-02-54


kolunmi:
Awesome! Thanks!

PR #1769, opened 2026-07-29
Workaround cache invalidation issue for favorites

AlexanderVanhee:
This should probably be fixed on Flathubs end, but this a fix that does technically work. https://github.com/flathub-infra/website/issues/6898
image


kolunmi:
bazaar was originally named gnome-apps-next :sob:

AlexanderVanhee:
Random lore drop

PR #1775, opened 2026-07-30
Fix plural index issue for "Trashed user data" string

AlexanderVanhee:
This is needed for languages like Polish or Arabic where there are more than 2 forms.


kolunmi:
Thank you

PR #1780, opened 2026-07-31
Request auto start when turning on auto-updates

AlexanderVanhee:
Makes a dbus call to the background portal to request auto-start when turning on auto-updates, this creates a .desktop file in `~/.config/autostart` to start Bazaar with the  `--no-window` parameter.

Before this, the daemon would only start up when first searching in the shell or when first running the actual app, so no auto-updates would happen until then.

### DIFF
--- a/src/bz-preferences-dialog.blp
+++ b/src/bz-preferences-dialog.blp
@@ -56,7 +56,9 @@ template $BzPreferencesDialog: Adw.PreferencesDialog {
         activatable-widget: automatic_updates_check;
 
         [prefix]
-        CheckButton automatic_updates_check {}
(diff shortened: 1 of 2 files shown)


kolunmi:
Thanks! We need to use libdex for the dbus stuff:

https://gnome.pages.gitlab.gnome.org/libdex/libdex-1/func.bus_get.html
https://gnome.pages.gitlab.gnome.org/libdex/libdex-1/func.dbus_connection_call.html

AlexanderVanhee:
Should be better now

kolunmi:
Perfect, thanks

PR #1781, opened 2026-07-31
Have separate version for cache wipes

AlexanderVanhee:
Makes it so we don't have to wipe the cache for smaller releases that don't touch any cached data.

It also allows us to reset the cache of other developers without having to do a version bump.

Also fixes an issue where the cache would be wiped, but not be refreshed if the user already refreshed the cache in the last day.

Closes #1755


kolunmi:
Looks good, Thanks!!

PR #1786, opened 2026-08-03
Add close button to donation banner

AlexanderVanhee:
This does require you to inline the Banner widget, which is pretty annoying.

Screenshot From 2026-08-03 14-25-00

Closes #1720


kolunmi:
I don't like this at all

PR #1789, opened 2026-08-04
Fix use after free in user data page

AlexanderVanhee:
Fixes an issue where the title of a fake group gets used for the toast after is has been freed.


kolunmi:
Thank you

PR #1790, opened 2026-08-04
Make read only apps appear in the installed apps list when loaded from groups cache

AlexanderVanhee:
This fixes an issue where the Bazaar app would not appear in the list until after a refresh happened.


kolunmi:
Thank you

PR #1792, opened 2026-08-04
Move auto-updates notice out of a menu button

### LINKED ISSUE #547: Provide options to disable automatic update checking
It would be good to expose some options that limit Bazaar’s ability to check for updates automatically instead of always doing so by default. Having these available is good (and arguably necessary) for users in low bandwidth or on metered connection (which is a lot of people) as well as considerate of power usage (since internet usage consumes power).

Image

These options may not necessarily match the concept above, but reflect however the automatic functionality of Bazaar could be limited.

AlexanderVanhee:
Closes #547

image


kolunmi:
Thank you

PR #1794, opened 2026-08-04
Clarify affected ubuntu versions in warning dialog

renner0e:
Ubuntu 24.04 is still being supported and used. It does not suffer from https://github.com/bazaar-org/bazaar/issues/583.

I think it is still worth showing the warning to unaffected 24.04 users because in the future Bazaar will stop working as soon as they upgrade their system.

Fixes: https://github.com/bazaar-org/bazaar/issues/1708
Screenshot_20260805_013935


kolunmi:
Thank you

PR #1795, opened 2026-08-05
Handle background transactions better

AlexanderVanhee:
Keeps the main application running until all transactions are complete, and sends a notification if all windows are closed when an application finishes installing, making sure the user stays informed.

image

Closes #1793

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -150,4 +150,5 @@ src/io.c
 src/main.c
 src/safety-calculator.c
 src/shortcuts-dialog.blp
+src/transaction-notification.c
 src/update-worker.c
--- a/src/bz-application.c
+++ b/src/bz-application.c
@@ -312,6 +312,11 @@ disable_blocklists_changed (BzApplication *self,
                             BzStateInfo   *state);
 
 static void
+transactions_active_changed (BzApplication        *self,
+                             GParamSpec           *pspec,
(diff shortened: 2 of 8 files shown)


kolunmi:
Thanks, IO should be in a fiber

kolunmi:
also formatting

AlexanderVanhee:
Does it matter here? It can't really profit from parallelization and blocking the main thread shouldn't really matter as there are no windows active.

kolunmi:
It's part of the style requirements.

kolunmi:
Also what is your reasoning for making a new `transaction-notification.c` file? Why didn't you use the existing `bz-transaction.c`?

AlexanderVanhee:
Because bz-transaction is just a data object that does not depend on GTK or anything, I originally had it in transaction manager but having it directly depend on things like bz-async-texture is kinda weird. I might also want to add things like progress notifications later, if it ever gets added to GNOME Shell or whatever.

kolunmi:
Ok, sounds good

AlexanderVanhee:
Changed it to use a fiber

kolunmi:
> Changed it to use a fiber

I meant in async-teture.c, where you called g_file_load_bytes

Sorry, I wasn't clear

kolunmi:
Thank you

PR #1796, opened 2026-08-05
Use both CACHE and PACKAGE versions

AlexanderVanhee:
I kinda made it so the donation prompt doesn't appear until the cache version got changed. What I should have done is use both values, the package version for the banner and the cache version for cache invalidation.


kolunmi:
Thanks

PR #1797, opened 2026-08-05
Only show message for effected Ubuntu versions

AlexanderVanhee:
Previously all Ubuntu versions got this warning, now only the effected ones ( =< 25.10).


kolunmi:
Thanks

PR #1807, opened 2026-08-06
Add stale updates notification via the update-worker

AlexanderVanhee:
Adds a notification when it has been more than 7 days since the last successful update check done either by the auto-update system or just normally in the app. The notification should only show once every 24 hours.


kolunmi:
Thank you

PR #1808, opened 2026-08-06
Use pt for Portuguese (Portugal) translation

haaninjo:
The [GNOME translation team for Portuguese](https://l10n.gnome.org/teams/pt/) uses pt rather than pt_PT for the Portugal variant. Change code so the team can start from the existing translation. Bazaar would be the only module with a pt_PT translation.

Thanks to pt_BR coordinator @rffontenelle for confirming my idea of way going forward.

Tested and confirmed to show same result when running with `LANG=pt_PT.UTF-8 bazaar` as it falls back to pt.


kolunmi:
Cool, thanks!

PR #1809, opened 2026-08-06
chore: pin flit_core to 3.X

renner0e:
Follow-up of b74b4ff.

Pyparsing can't currently be built with 4.X.

https://github.com/pyparsing/pyparsing/blob/86fac1d6313de4cd8558179e8599c9eb53c272ca/pyproject.toml#L2


kolunmi:
Thanks

PR #1812, opened 2026-08-07
Make dialogs more accessible

AlexanderVanhee:
Makes the titles under the lozenges get read by screen readers in dialogs where it makes sense. Also ensures that all text on the arbitrary permissions page is read before the button.


kolunmi:
Thanks

PR #1823, opened 2026-08-09
Add Flathub's curated sections to the home page

AlexanderVanhee:
This PR implements the curated sections found on the Flathub page. There are a total of 3 slots where they can appear, currently only two are being used.
There are also 13 theme keys that bring their own title and subtitle. If Flathub decides to add a different theme key that we don't support, then the title and subtitle get hidden but the apps will still appear.
I slightly changed where the slots are placed because the app of the day is in a different location. I also limited the apps to 8 or 6 so it would look good on our 4 wide layout.

image

image


AlexanderVanhee:
@razzeee is the API or design of this system subject to change by much?

razzeee:
I expect strings, specifically descriptions to change slightly. 

We might also add another slot as after first category feels too early. So likely after second category, wasn't planning on doing that in a non braking way, but we can. 

Otherwise nothing else is planned, we might add some fields, if we decide to have more looks or graphics. But that should be non breaking.

kolunmi:
Thanks

PR #1830, opened 2026-08-11
transaction tile: Make progress label numeric

YaLTeR:
This label has a frequently updating number (file size), and the numeric class makes it monospace, so that it doesn't jump around. This is a libadwaita built-in class.

Not sure off the bat if any other labels similarly need this.


kolunmi:
Thank you

PR #1831, opened 2026-08-11
Update string for Exhibit app page screenshot

haaninjo:
Multiple translations took it as the verb "exhibit"
rather than an app name. Reorder so it's clearer that it's not the verb.


kolunmi:
Thank you

PR #1832, opened 2026-08-11
feat: only track stable flatpak version

renner0e:
1.19 was just tagged, we should stick with the stable release series.

We can't use anitya for this because the 1.17 odd release series were not marked as pre-release :/


renner0e:
flathub: https://github.com/flathub/io.github.kolunmi.Bazaar/pull/149

kolunmi:
Thank you

PR #1840, opened 2026-08-13
Fix outline of radio buttons in preferences

AlexanderVanhee:
Had this exact same issue in the transaction dialog, and still did not learn.


kolunmi:
Thanks

PR #1841, opened 2026-08-13
Fix doap syntax

haaninjo:
l10n.gnome.org requires mbox to be on same level as name of the maintainer

Hopefully enough to get Damned Lies to pick up maintainer this time. At least I got it to work in a VM after doing the same change:

`damnedlies - INFO - retrieve_or_create_person in __init__: Could not retrieve person <DOAPMaintainer: Anders (mail: anders@example.com; account; None)>. This maintainer will be created in the system.`


kolunmi:
Thanks!

PR #1842, opened 2026-08-14
Increase padding of OSD elements on screenshot page

AlexanderVanhee:
An increase is needed to stop the scroll bars from overlapping.


kolunmi:
Thanks

PR #1845, opened 2026-08-14
Add zoom shortcuts

AlexanderVanhee:
https://developer.gnome.org/hig/reference/keyboard.html#view-options Implements the applicable shortcuts seen here.


kolunmi:
We should probably use a shortcut controller for this, and bind to widget actions...

AlexanderVanhee:
Updated it to use actions and a shortcut controller in the blp

kolunmi:
Thanks

PR #1847, opened 2026-08-14
Recreate full view on each navigation to fix anim

AlexanderVanhee:
This fixes the animation issue by creating a new page with the full view every time the user clicks an app.

To keep RAM usage low, the actual full view widget in the page gets unloaded when the page it's in is hidden.

This approach does come with the cost of needing to create a new full view widget every time, but the overhead seems pretty small, on my system it was 11ms the first time and 7ms for subsequent creations.


kolunmi:
Thanks

PR #1850, opened 2026-08-15
Add drop support for .flatpak and .flatpakrefs

AlexanderVanhee:
Adds drop support, while making the icon change when dragging a file over the app.


kolunmi:
Nice, thanks

PR #1855, opened 2026-08-16
Fix metainfo file to support mobile

AlexanderVanhee:
display_length seemingly has to be in "requires" if you look at this document: https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#control


kolunmi:
Thanks

PR #1863, opened 2026-08-16
chore: refined strings + renamed icons + adwaita category change

theMimolet:
Good evening :D

This is a proof of concept of string changes I suggested
and wondered if you'd be okay with them or not.

Here are the changes:

## ~~"Stopped Receiving Core Updates" / "Stopped Receiving Updates"~~

~~I replaced them by "Uses an End-of-Life runtime" as it's more accurate and is more informative (because we clearly say what's up with those).~~ 

~~Some apps - mostly outside Flathub - can still be actively developed but use EOL runtimes. We agree that it's not a great practice, but it happens...~~

~~(I believe some up-to-date versions of Polytoria used to have EOL runtimes for example)~~ 

**Update**: The slight misalignment incentivizes app devs using an updated runtime

image

## "Ubuntu Troubles!"

I softened the title ("Notice for Ubuntu Users") and the notice itself : 
> "Ubuntu 25.10 and newer have changed default security rules in a way that makes it <b>impossible to install apps</b> from the Flatpak version of Bazaar. For the time being, please use the apt package instead. Install it with
```sudo apt install bazaar```
You can then remove this Flatpak version with
```flatpak uninstall io.github.kolunmi.Bazaar```"

Less blaming, less stress, attempt to be more informative.

## "Malformed Link"

In this PR, I rephrased the message as the current one is a bit negative and could leave the user wondering if they did something wrong. So I replaced it by saying what Bazaar is currently doing : 
"Corrected Link"

>"The link used to open this app has incorrect capitalization and has been corrected by Bazaar.
This is most likely caused by KRunner sending incorrect app IDs"

But after thinking, it would be better to remove the message itself as it doesn't serve any real purpose to the user : the user wanted to access an app's  page, they access the page - they don't care Bazaar corrected the link, because on their end Bazaar did its job opening the right page.

## "Other apps"

It will need to be reviewed by a native english speaker, because it's just reorganising the strings for "Other apps from the same dev" and such. It also includes Alexander's request to rename the "Other apps by dev" to a "All apps from dev" page as it does exactly show that (all the apps from the same dev). 

I'm just not sure on the wording. 

image

image

## ~~"Adwaita" category~~

~~Might be the most impactful and controversial.~~

~~The "Adwaita" category name might not fit for a large target audience, because the only people that know - and should know - Adwaita are nerds.~~

~~For most people, they know their Ubuntu, their Fedora, their Bluefin (because it's the name of their system), maybe they know about Gnome (because it's the name of their desktop and it's shown a bit everywhere), but I doubt people like my mother would know what Adwaita is if I asked her.~~

~~As an attempt, I rebranded the Adwaita category into a "GNOME Apps" category, just like KDE.
I know it's not a perfect solution, since I believe it's also shipped to other desktops, but I had to try it.~~

image

**Update**: 
- "I suppose it’s a good learning opportunity, if they don’t they’ll just click the section and learn *oh, adwaita is a kind of app that has that certain look* Adwaita itself isn’t really a technical term"
- "Longer term i think i'd rather just move this to the home page somehow, and have it go via the flathub content pipeline and then brand it "Built for your system" or something but for now this is fine"

Also while replacing the Adwaita icon by Gnome's, I noticed there was like  "io.github.kolumni.Bazaar.Kde.svg" and "io.github.kolunmi.Bazaar.kde.svg" which is awful naming. So I renamed the icons used on the login screen with a "login-".

## Um

Once again, this is a proof of concept, some might need discussion, some might get dropped and I expect you recheck it thoroughly before it gets merged (if it gets merged).

PS : Also doesn't Source and Remote mean the same thing ? 🥲

### CODE REVIEW

File: src/bz-addon-tile.blp [hunk 1]
```diff
@@ -61,7 +61,7 @@ template $BzAddonTile: $BzListTile {
         single-line-mode: true;
         halign: start;
         hexpand: true;
-        label: _("Stopped Receiving Updates");
+        label: _("Uses an End-of-Life runtime");
```
renner0e: This is the correct term but will non-technical users understand what this means? What is a runtime? What does end-of-life mean?

theMimolet: You're right, however both those terms are already used all over Bazaar.

Moreover, there's a description of what an End-of-Life app is in the settings and there's a description of what problems an End-of-Life runtime could give the user when they open the app page.  

A solution could be to add a precision on what a runtime is in the EOL runtime description.

bertob:
Mostly looks fine now, some minor notes:

> But after thinking, it would be better to remove the message itself as it doesn't serve any real purpose to the user : the user wanted to access an app's page, they access the page - they don't care Bazaar corrected the link, because on their end Bazaar did its job opening the right page.

+1, if the link breaks in the future that's a future problem (and not actionable for non-developers anyway). I wouldn't annoy everyone who clicks the link about it.

> It will need to be reviewed by a native english speaker, because it's just reorganising the strings for "Other apps from the same dev" and such. It also includes Alexander's request to rename the "Other apps by dev" to a "All apps from dev" page as it does exactly show that (all the apps from the same dev).

Would be nice to avoid the super long label on the button, I'd go with something simpler like "Show All" for that since the developer name is already in the section title.

